PR #1, opened 2017-05-18
port over and have nats transport

rybit:
This should be very compatible with the existing lib. We just swap it into place and configure the different transports.

I will be adding a signalFX and datadog transport too


calavera:
LGTM

PR #4, opened 2017-06-28
Add dimensions support to stats

brycekahle:
Also change to use cumulative counters for SignalFX.

### DIFF
--- a/metrics/transport/signalfx.go
+++ b/metrics/transport/signalfx.go
@@ -52,5 +52,14 @@ func (t *SFXTransport) Publish(m *metrics.RawMetric) error {
 		p.Dimensions[k] = asStr
 	}
 
+	switch m.Type {
+	case metrics.GaugeType:
+		p.MetricType = datapoint.Gauge
+	case metrics.TimerType:
+		fallthrough
+	case metrics.CounterType:
+		p.MetricType = datapoint.Counter
+	}
+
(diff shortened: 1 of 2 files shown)


metrics/transport/signalfx.go
```diff
@@ -52,5 +52,14 @@ func (t *SFXTransport) Publish(m *metrics.RawMetric) error {
 		p.Dimensions[k] = asStr
 	}
 
+	switch m.Type {
+	case metrics.GaugeType:
+		p.MetricType = datapoint.Gauge
+	case metrics.TimerType:
+		fallthrough
```
rybit: Timers should be reported as timers right?

brycekahle: SignalFX doesn't have timers 😄

rybit: (╯°□°）╯︵ ┻━┻

I swear they said they did...but you're right.

rybit:
lgtm

brycekahle:
Any concerns about running a SHA256 every time a metric is recorded?

rybit:
So I did some benchmarking on this before I added it. I am not worried. 

```
BenchmarkSignature-8         300000        4052 ns/op      1216 B/op        27 allocs/op
```

meaning that we could do around 250k signatures/second.

PR #5, opened 2017-06-29
refactor the stats package to make better experience

rybit:
The stats package was going to put out a lot of lines potentially, this means it will be only one. And it makes logging at all optional.

Also a test.

### DIFF
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -76,8 +76,11 @@ func report(log *logrus.Entry, prefix string) {
 			name += k
 			metrics.CountN(name, val, dims)
 
-			dims["value"] = val
-			results[name] = append(results[name], dims)
+			resMap := map[string]interface{}{
+				"value": val,
+				"dims":  dims,
+			}
+			results[name] = append(results[name], resMap)
 		}
 	}
 	statLock.Unlock()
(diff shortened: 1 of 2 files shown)


stats/stats.go
```diff
@@ -29,45 +29,67 @@ var mtsMap = map[string]map[string]int64{}
...
+		for sha, val := range series {
+			dims := dimMap[sha]
+			name := prefix
+			if name != "" && !strings.HasSuffix(name, ".") {
+				name += "."
+			}
+			name += k
+			metrics.CountN(name, val, dims)
+
+			dims["value"] = val
```
brycekahle: If you have two different metrics with the same dimensions, wouldn't setting `value` here clobber the previous values?

rybit: The only time this would clobber is that if there was a "value" dimension. The `dims` here is per MTS

brycekahle: The `dimsMap` is keyed by a SHA256 of the dims without the metric name though. If you have two metrics, with the same dims, they will have the same SHA, and resolve to the same entry in `dimsMap`.

A simpler example would be *no dims*. If you had two metrics with no dims and different values, then the log message would use the same `value`.

Would it be easier if I added a failing test?

rybit: Ah I see what you mean. But that is ok. I can actually just refactor what is stored there.

brycekahle:
:shipit:

PR #6, opened 2017-06-30
Do not log stats if there are none

brycekahle:


rybit:
This will work, but I think that the clearer code would be to check len(mtsMap) at the entry of the method and then just exit.

rybit:
shipit

PR #7, opened 2017-06-30
unify the metrics and stats packages

rybit:
it was really unnecessary to have two packages. It was born of laziness. This will simplify that.


metrics/reporter.go
```diff
@@ -0,0 +1,112 @@
...
+	if atomic.SwapUint32(&r.shutdownFlag, running) == running {
+		// going from running -> running
+		return
+	}
+
+	go func() {
+		if r.log != nil {
+			r.log.WithFields(logrus.Fields{
+				"interval": r.interval,
+			}).Infof("Starting to report stats every %d seconds", r.interval)
```
brycekahle: This will print the interval in nanoseconds, won't it?

metrics/reporter.go
```diff
@@ -0,0 +1,112 @@
...
+
+	results := make(map[string][]map[string]interface{})
+
+	for _, c := range r.counters {
+		for _, m := range c.series() {
+			err := m.send(nil)
+			if err != nil {
+				r.log.WithError(err).Warnf("Failed to send %+v", m)
+			}
+		}
```
brycekahle: Shouldn't this loop be adding entries to `results`?

metrics/reporter.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (r *intervalReporter) report() {
+	if len(r.counters) == 0 {
+		return
+	}
+
+	results := make(map[string][]map[string]interface{})
+
+	for _, c := range r.counters {
```
brycekahle: Do we need to use the metric mutex at all when obtaining the values for reporting?

metrics/reporter.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}()
+}
+
+func (r *intervalReporter) stop() {
+	if atomic.SwapUint32(&r.shutdownFlag, stopped) == stopped {
+		// going from stopped -> stopped
+		return
+	}
+	close(r.shutdown)
```
brycekahle: Closing the channel will prevent the reporter from ever starting up again. Is that intentional?

rybit: Yes. If the threads relying on this (consuming from shutdown) crashed, this would hang the call to shutdown. In this way, we can broadcast a shutdown. It is all private to the lib too, you'd just want to make a new reporter

metrics/reporter.go
rybit: I was thinking about this. I don't think it is strictly necessary because what would happen is that a value would be added or modified during the loop. This isn't a big deal because we'd report the higher value the next time.

brycekahle:
🚢

PR #14, opened 2017-07-19
create the graceful server package

rybit:
This is derived heavily from https://github.com/netlify/gocommerce/pull/65/files

It is intentionally split out into Bind and Listen methods (with a ListenAndServe) so that testing is easier. It also intentionally uses the listener directly because it supports using 0 as a port and letting the system decide the port.

### DIFF
--- a/graceful/graceful.go
+++ b/graceful/graceful.go
@@ -1,0 +1,86 @@
+package graceful
+
+import (
+	"context"
+	"errors"
+	"net"
+	"net/http"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+var DefaultShutdownTimeout = time.Second * 60
+
+type GracefulServer struct {
+	server   *http.Server
+	listener net.Listener
+	log      *logrus.Entry
+
+	URL             string
+	ShutdownTimeout time.Duration
+}
+
+func NewGracefulServer(handler http.Handler, log *logrus.Entry) *GracefulServer {
+	return &GracefulServer{
+		server:          &http.Server{Handler: handler},
+		log:             log,
+		listener:        nil,
+		ShutdownTimeout: DefaultShutdownTimeout,
+	}
+}
+
+func (svr *GracefulServer) Bind(addr string) error {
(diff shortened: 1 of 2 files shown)


graceful/graceful.go
```diff
@@ -0,0 +1,86 @@
...
+		return errors.New("The listener has already started, don't call Bind first")
+	}
+	if err := svr.Bind(addr); err != nil {
+		return err
+	}
+
+	return svr.Listen()
+}
+
+func (svr *GracefulServer) Close() error {
```
brycekahle: Can we rename this to `Shutdown`? It gets confusing when there is also a `Close` on `http.Server`.

graceful/graceful.go
```diff
@@ -0,0 +1,86 @@
...
+		return err
+	}
+	svr.URL = "http://" + l.Addr().String()
+	svr.listener = l
+	return nil
+}
+
+func (svr *GracefulServer) Listen() error {
+	go svr.listenForShutdownSignal()
+	return svr.server.Serve(svr.listener)
```
brycekahle: Remember that this will return as soon as `http.Server.Shutdown()` is called. Should we do anything in this package to make that easier, and not something the user has to remember? I worry about calls to `os.Exit` afterwards. `listenForShutdownSignal` is the method that blocks until the server shutdown is complete or times out.

rybit: hmmm good point. I forgot about that.

graceful/graceful.go
rybit: I was following test server...https://golang.org/pkg/net/http/httptest/#Server.Close 

But I am totally fine with changing it.

brycekahle: Wow, way to be confusing go. https://golang.org/pkg/net/http/#Server.Close

> Close immediately closes all active net.Listeners and any connections in state StateNew, StateActive, or StateIdle. For a graceful shutdown, use Shutdown.

brycekahle:
I noticed we don't have CI running on this repo. Should we fix that?

rybit:
yes! I'll copy over the necessary files

PR #16, opened 2017-07-20
create a toml package

rybit:


ntoml/netlify_toml.go
```diff
@@ -0,0 +1,103 @@
...
+	"path"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+	"github.com/pkg/errors"
+)
+
+const DefaultFilename = "netlify.toml"
+
+type NetlifyToml struct {
```
calavera: This doesn't match the configuration in buildbot. Let's make sure they are compatible:

https://github.com/netlify/buildbot/blob/master/bot/configuration/configuration.go#L19-L31

rybit: should be addressed

calavera:
LGTM

brycekahle:
What is the benefit of having the configuration definitions here instead of in the projects that use them?

rybit:
it is because we don't have a place to specify "what can you set" in the toml file. It leads to overlapping configuration and "I think you can do that...". It also doesn't let us define some common structures (e.g. settings).

ntoml/netlify_toml.go
```diff
@@ -0,0 +1,111 @@
...
+
+func LoadFrom(paths ...string) (*NetlifyToml, error) {
+	if len(paths) == 0 {
+		return nil, errors.New("No paths specified")
+	}
+
+	out := new(NetlifyToml)
+
+	for _, p := range paths {
+		if _, err := os.Stat(p); err == nil {
```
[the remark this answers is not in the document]
brycekahle: The reason for this is that anything can happen on the filesystem between the `Stat` and `ReadFile` calls. If you want to ignore paths that don't exist, best to do it on `ReadFile`.

ntoml/netlify_toml.go
```diff
@@ -0,0 +1,111 @@
...
+			data, ferr := ioutil.ReadFile(p)
+			if ferr != nil {
+				return nil, errors.Wrapf(ferr, "Error while reading in file %s.", p)
+			}
+
+			if _, derr := toml.Decode(string(data), out); derr != nil {
+				return nil, errors.Wrapf(err, "Error while decoding file %s", p)
```
brycekahle: should be `derr` instead of `err`.

ntoml/netlify_toml.go
```diff
@@ -71,14 +71,14 @@ func LoadFrom(paths ...string) (*NetlifyToml, error) {
...
-			data, ferr := ioutil.ReadFile(p)
-			if ferr != nil {
-				return nil, errors.Wrapf(ferr, "Error while reading in file %s.", p)
-			}
+		data, ferr := ioutil.ReadFile(p)
+		if ferr != nil {
+			return nil, errors.Wrapf(ferr, "Error while reading in file %s.", p)
+		}
 
+		if !os.IsNotExist(ferr) {
```
brycekahle: `ferr` will always be `nil` here. You need to check `os.IsNotExist` first.

PR #17, opened 2017-07-24
Split Rabbit initialization to be more flexible.

calavera:
- Create DialToRabbit function to get only a connection.
- Allow to create multiple consumers based on a connection.

Signed-off-by: David Calavera <david.calavera@gmail.com>


rybit:
lgtm

PR #18, opened 2017-07-24
close a channel to indicate the server is closing

rybit:
We were having the case that this would timeout because we would be waiting for the exit signal on shutdown and the listen to exit. This fixes that I think.


calavera:
LGTM

PR #24, opened 2017-08-10
set TLS values directly

rybit:
In this case we allow for using the values directly, not from file. It is good for kube services.

Also, if no CA is specified we will fall back to the system ones.


calavera:
LGTM

PR #26, opened 2017-08-15
add support for reflecting into missing stucts

rybit:
This won't support pointers to strings/int/etc


brycekahle:
Should we add tests for env vars since that is what we are concerned about breaking?

rybit:
good call @brycekahle, added!

PR #28, opened 2017-08-16
migrate to using env config

rybit:
This should let us configure from the environment, not files, more easily.

We can still use files from the environment. When updating to this version of the lib though, we would not support json/toml/yaml any longer.

I think that this makes sense (after some convincing from @brycekahle ) because many of our services are moving to kubernetes and that will make it much easier to work in that environment. It _should_ simplify the templates too.

It will required that we change all the tags on structures when the service is moved over.


rybit:
an open question I have is how it handles things like a `map[string]struct{}`? Or `[]struct{}`. 


Those are really common

rybit:
or doing arbitrary k/v maps. That doesn't seem supported easily.

rybit:
The more I think on this, the less I think this is going to work. Yes, it is good to be able to configure out of env vars. But outside the simple use cases, some lightly nested objects, this falls apart. 

For concrete instances:
  - [arbitrary maps](https://github.com/netlify/netlify-commons/blob/master/nconf/metrics.go#L29)
  - [simple structures](https://github.com/netlify/ns1controller/blob/master/conf/configuration.go#L30)
  - https://github.com/netlify/buildbot/blob/master/consumer/messages/cmd_data.go#L8
  - [maps to structs](https://github.com/netlify/proxy/blob/master/conf/configuration.go#L11)

I am betting many of those can be solved using [custom encoders](https://github.com/netlify/proxy/blob/master/conf/configuration.go#L11). But that will mean we will have to have some convention for serializing those to env vars. And then there is the issue of hitting the practical limit of the max length of environment variables.

rybit:
It might be that we'd then have to have some dependency on a file for _some_ config and then on env vars for others. I don't like that

brycekahle:
maps to non-struct types:
> export MYAPP_COLORCODES="red:1,green:2,blue:3"

slice of structs: These wouldn't have worked from the environment with viper either. I don't think viper has support for slices from env vars at all, does it?

Realistically, I think we need to strive for environment variable configuration for docker apps where possible. That may mean restructuring the configuration. If not possible, then we will be doing things like [this](https://github.com/netlify/ansible/blob/9370229dea13a1e786b1903a7fdc1ef085e611e5/kubernetes/log-watcher/kubernetes-spec.yml#L33-L41).

For non-docker apps, I think we can use whatever we like.

brycekahle:
We may also start loading configuration from consul in the near future too.

rybit:
So yes it does support maps of structs, but look at the ns1controller config or the proxy config, those end up being really cumbersome to understand when jammed into an env var. 

I am not sure on the slice of structs, I know that it would work from file but not sure on the env var. It seems very against the way that envvars are meant to be used. 

And to address consul & docker directly:
  - yes, we may one day directly read from consul. But all the stuff I have read has recommended using a buffer (consul-templates) for production services. This means that you can still operate if the consul cluster goes down. So directly reading from consul? That doesn't seem too likely
  - docker, yes we def need to handle that better. For instance, it would be very reasonable to put the ns1controller in there. I don't have a good idea on _how_ we would load the more complex config. We could do something nasty like serialize a bunch of json into an envvar...but please no. I am def open to suggestions. 

And for perspective, the ns1controller has a reasonably small config and it is 5k characters in production. The limit on the env is around 32k ~ for all the environment. Just something to consider.

I am open to options about how to better do configuration

rybit:
ok not going to do this. It is conflating too many issues. We are instead going to push forward with _no_ env support in commons and tackle it as we need it

PR #31, opened 2017-09-04
[tls] Fix CA cert conditional

sgirones:
With the previous code, it uses the system certificates if a CA cert is provided


calavera:
LGTM

PR #32, opened 2017-09-08
Add derivative timing metrics to signalFX

brycekahle:


metrics/transport/signalfx.go
```diff
@@ -58,10 +79,57 @@ func (t *SFXTransport) Publish(m *metrics.RawMetric) error {
 	case metrics.CumulativeType:
 		p.MetricType = datapoint.Counter
 	case metrics.TimerType:
-		fallthrough
+		if t.reportDelay > 0 {
+			if err := t.recordTimer(m, p.Dimensions); err != nil {
+				return errors.Wrap(err, "error recording timer to histogram")
+			}
+		}
+		p.MetricType = datapoint.Count
```
rybit: do we want to be still recording these in a bucket?

brycekahle: I did this to be 100% backwards compatible until we are happy with the new metrics.

rybit:
One thing I have been thinking about is that we could expose these as a different datatype. We would report each timer directly (useful for when we want to track _actual_ times).

Maybe something like a histogram type?

brycekahle:
I thought about that, but this allows us to get all the derivative metrics without changing any app code.

rybit:
we are going to have to change the code to upgrade to this lib. It should be a quick change, and would be a bit clearer. I'd even just take some kind of flag or something on the metric? (not in love with that idea)

brycekahle:
Since we are reporting derivative metrics only for SFX, it also didn't make sense to make a new type. What would the new type do for the other transports?

rybit:
We could have it always report the rollups on some interval. It would operate in much the same way as the cumulative counter

rybit:
One of the uses I was thinking about this was in buildbot vs netlify-server. In the former we want to report _each_ timing, the rollups would be ok, but the data is sooo low volume that it makes sense to do each one. And then we can chart things like "how long did this particular site take". 

Whereas, in netlify-server we want to know in a more aggregate fashion what a block's timing is.

brycekahle:
> And then we can chart things like "how long did this particular site take"

We can do that from the logs now. In order to get that functionality in SFX, we would need to include the site as a dimension, which would have unbounded cardinality. Isn't that generally a bad idea?

PR #33, opened 2017-09-15
support env configuration

rybit:
We are moving to support environmental configuration, this adds some tags and cleanup to make that easier. We can't do nil values with env configs, so we need to rely on empty strings or enabled flags

### DIFF
--- a/nconf/metrics.go
+++ b/nconf/metrics.go
@@ -19,11 +19,7 @@ type MetricsConfig struct {
 
 	SFXToken string `mapstructure:"sfx_token"`
 
-	Nats *struct {
-		TLS     *tls.Config `mapstructure:"tls_conf"`
-		Servers []string    `mapstructure:"servers"`
-		Subject string      `mapstructure:"subject"`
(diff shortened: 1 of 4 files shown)


calavera:
LGTM

nconf/metrics.go
```diff
@@ -13,23 +13,25 @@ import (
...
 type MetricsConfig struct {
 	DataDog *struct {
-		APIKey string `mapstructure:"api_key"`
-		AppKey string `mapstructure:"app_key"`
+		APIKey string `mapstructure:"api_key" split_words:"true"`
+		AppKey string `mapstructure:"app_key" split_words:"true"`
 	} `mapstructure:"datadog"`
 
-	SFXToken string `mapstructure:"sfx_token"`
+	SFXToken string `mapstructure:"sfx_token" split_words:"true"`
```
brycekahle: I'm not 100% sure what will happen with `split_words` here. It might be `S_F_X_TOKEN` 😅 .

PR #34, opened 2017-09-18
use travis ci

rybit:


.travis.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+language: go
+
+go:
+  - 1.8
```
brycekahle: I think this uses 1.8.0 instead of 1.8.3

.travis.yml
```diff
@@ -0,0 +1,12 @@
...
+language: go
+
+go:
+  - 1.8.3
+
+install: make deps
+script: make test
+
+notifications:
+  email: false
```
brycekahle: Why no emails? A user can turn them off if they don't want them.

rybit: b/c I copied this from gotrue 😆

calavera: they are very noisy

.travis.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+language: go
+
+go:
+  - 1.8.3
```
calavera: let's add `1.9.x` to this list, so we build in both. 1.9 is the current stable.

PR #35, opened 2017-09-19
support for optional TLS in rabbit and nats

rybit:
The TLS configuration can be nil, meaning we don't want TLS. We need to respect that. Also adds some debugging messages to the rabbitMQ connection code and the envconfig tags


calavera:
LGTM

PR #38, opened 2017-10-02
[discovery] Add DNS discovery methods. Rabbit and Nats config to support DNS discovery

sgirones:
* Get a list of hosts from a DNS name
* Use a DNS name for configuring Rabbit and Nats

### DIFF
--- a/discovery/main.go
+++ b/discovery/main.go
@@ -1,9 +1,7 @@
 package discovery
 
 import (
-	"fmt"
-
-	"github.com/miekg/dns"
+	"net"
 )
 
 type Endpoint struct {
@@ -13,50 +11,18 @@ type Endpoint struct {
 
 func DiscoverEndpoints(serviceDNS string) ([]Endpoint, error) {
 	endpoints := []Endpoint{}
-	qType := dns.StringToType["SRV"]
-	serviceDNS = dns.Fqdn(serviceDNS)
+	_, remotes, err := net.LookupSRV("", "", serviceDNS)
 
-	client := &dns.Client{}
-	msg := &dns.Msg{}
-
-	msg.SetQuestion(serviceDNS, qType)
-
-	dnsserver, err := getDNSServer()
 	if err != nil {
 		return endpoints, err
 	}
 
-	response, _, err := client.Exchange(msg, dnsserver)
-	if err != nil {
-		return endpoints, err
-	}
-
-	if msg.Id != response.Id {
-		return nil, fmt.Errorf("DNS ID mismatch, request: %d, response: %d", msg.Id, response.Id)
-	}
-
-	for _, v := range response.Answer {
-		if srv, ok := v.(*dns.SRV); ok {
-			endpoints = append(endpoints, Endpoint{
-				Name: srv.Target,
-				Port: srv.Port,
-			})
-		}
+	for _, n := range remotes {
+		endpoints = append(endpoints, Endpoint{
+			Name: n.Target,
+			Port: n.Port,
+		})
 	}
 
 	return endpoints, nil
-}
-
-func getDNSServer() (string, error) {
-	servers, err := parseResolvConf()
-	return servers[0], err
-}
-
-func parseResolvConf() ([]string, error) {
-	conf, err := dns.ClientConfigFromFile("/etc/resolv.conf")
-	if err != nil {
-		return []string{}, err
-	}
-
-	return conf.Servers, nil
 }
--- a/discovery/main.go
+++ b/discovery/main.go
@@ -1,0 +1,62 @@
+package discovery
+
+import (
+	"fmt"
+
+	"github.com/miekg/dns"
+)
+
+type Endpoint struct {
+	Name string
+	Port uint16
+}
+
+func DiscoverEndpoints(serviceDNS string) ([]Endpoint, error) {
+	endpoints := []Endpoint{}
+	qType := dns.StringToType["SRV"]
+	serviceDNS = dns.Fqdn(serviceDNS)
+
+	client := &dns.Client{}
+	msg := &dns.Msg{}
+
+	msg.SetQuestion(serviceDNS, qType)
+
+	dnsserver, err := getDNSServer()
+	if err != nil {
+		return endpoints, err
+	}
+
+	response, _, err := client.Exchange(msg, dnsserver)
+	if err != nil {
+		return endpoints, err
+	}
+
+	if msg.Id != response.Id {
+		return nil, fmt.Errorf("DNS ID mismatch, request: %d, response: %d", msg.Id, response.Id)
+	}
+
+	for _, v := range response.Answer {
+		if srv, ok := v.(*dns.SRV); ok {
+			endpoints = append(endpoints, Endpoint{
+				Name: srv.Target,
+				Port: srv.Port,
+			})
+		}
+	}
+
+	return endpoints, nil
+}
+
+func getDNSServer() (string, error) {
+	servers, err := parseResolvConf()
+	return servers[0], err
+}
+
+func parseResolvConf() ([]string, error) {
+	conf, err := dns.ClientConfigFromFile("/etc/resolv.conf")
+	if err != nil {
+		return []string{}, err
+	}
+
+	return conf.Servers, nil
+}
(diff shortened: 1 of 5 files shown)


discovery/main.go
```diff
@@ -0,0 +1,62 @@
+package discovery
+
+import (
+	"fmt"
+
+	"github.com/miekg/dns"
```
rybit: is there a reason to not use the builtin https://golang.org/pkg/net/#LookupSRV? 

We aren't specifying a particular DNS server, just using the resolve.conf, meaning we gain nothing afaict

messaging/nats.go
```diff
@@ -1,20 +1,23 @@
...
+	"github.com/netlify/netlify-commons/discovery"
 	"github.com/netlify/netlify-commons/tls"
 )
 
 type NatsConfig struct {
-	TLS         *tls.Config `mapstructure:"tls_conf"`
-	Servers     []string    `mapstructure:"servers"`
-	LogsSubject string      `mapstructure:"log_subject"`
+	TLS            *tls.Config `mapstructure:"tls_conf"`
+	ServiceDNSName string      `mapstructure:"service_dnsname"`
```
rybit: can you add an `envconfig:"dns_name"` so that we can access it by envvars easily? It is redundant to say "service_" b/c it is embedded in a nats config already

messaging/nats.go
```diff
@@ -64,6 +66,23 @@ func ConfigureNatsConnection(config *NatsConfig, log *logrus.Entry) (*nats.Conn,
 
 // ConnectToNats will do a TLS connection to the nats servers specified
 func ConnectToNats(config *NatsConfig, errHandler nats.ErrHandler) (*nats.Conn, error) {
+	serversString := config.ServerString()
+
+	if config.ServiceDNSName != "" {
```
rybit: we should pull this into a method.
```
if config.ServiceDNSName != "" {
  config.Servers = discoverNatsService(config.ServicDNSName)
}
```

discovery/main.go
sgirones: TBH I googled around and found that library to be used in some places. I will take a look at your suggestion

messaging/nats.go
sgirones: I wanted to add "service" to make it explicit that it was a "discovery" DNS name, not a simple DNS name pointing to a host

messaging/nats.go
sgirones: 👍

discovery/main.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"net"
+)
+
+type Endpoint struct {
+	Name string
+	Port uint16
+}
+
+func DiscoverEndpoints(serviceDNS string) ([]Endpoint, error) {
```
rybit: personally, I'd do something like this
```
func DiscoverEndpoints(serviceName string) ([]Endpoint, error) {
  _, remotes, err := net.LookupSRV("", "", serviceName) 
  if err != nil {
    return nil, err
  }

  endpoints := []Endpoint{}
  for _, n := range remotes {
    endpoints = append(...)
  }
  return endpoints, nil
}
```

a little different, mostly around some spacing and where I'd declare endpoints. That is mostly style though. 

I am not sure if we need the distinction though. I mean the remote structure is pretty close to the Endpoint in general. I don't think we are getting much by doing this in commons really.

messaging/nats.go
```diff
@@ -1,20 +1,23 @@
...
+	"github.com/netlify/netlify-commons/discovery"
 	"github.com/netlify/netlify-commons/tls"
 )
 
 type NatsConfig struct {
-	TLS         *tls.Config `mapstructure:"tls_conf"`
-	Servers     []string    `mapstructure:"servers"`
-	LogsSubject string      `mapstructure:"log_subject"`
+	TLS           *tls.Config `mapstructure:"tls_conf"`
+	DiscoveryName string      `mapstructure:"discovery_name"`
```
rybit: add a `split_words:"true"` so that the envconfig would be something like `WHATEVER_NATS_DISCOVERY_NAME`.

messaging/nats.go
```diff
@@ -92,3 +104,18 @@ func ErrorHandler(log *logrus.Entry) nats.ErrHandler {
 		}).Error("Error while consuming from " + sub.Subject)
 	}
 }
+
+func buildNatsServersString(serviceName string) (string, error) {
+	natsUrls := []string{}
```
rybit: /s/natsUrls/natsURLs/ ~ go uppercases any acronyms

messaging/nats.go
```diff
@@ -92,3 +104,18 @@ func ErrorHandler(log *logrus.Entry) nats.ErrHandler {
...
+	endpoints, err := discovery.DiscoverEndpoints(serviceName)
+	if err != nil {
+		return "", err
+	}
+
+	for _, endpoint := range endpoints {
+		natsUrls = append(natsUrls, fmt.Sprintf("nats://%s:%d", endpoint.Name, endpoint.Port))
+	}
+
+	return strings.Join(natsUrls, ","), nil
```
rybit: I would return the nats urls directly

messaging/nats.go
```diff
@@ -64,6 +66,16 @@ func ConfigureNatsConnection(config *NatsConfig, log *logrus.Entry) (*nats.Conn,
 
 // ConnectToNats will do a TLS connection to the nats servers specified
 func ConnectToNats(config *NatsConfig, errHandler nats.ErrHandler) (*nats.Conn, error) {
+	var err error
+	serversString := config.ServerString()
```
rybit: personally, I'd choose to do this after the discover block. Something like:
```
if config.DiscoveryName != "" {
  var err error
  config.Servers, err = discoverNatsURLs(config.DiscoveryName)
  if err != nil {
    return nil, errors.Wrapf(err, "Failed to lookup servers for service %s", config.DiscoveryName)
  }
}
...

return nats.Connect(config.ServerString(), options...)
```

messaging/rabbit.go
```diff
@@ -204,11 +206,20 @@ func ValidateRabbitConfigStruct(servers []string, exchange ExchangeDefinition, q
 
 // ConnectToRabbit will open a TLS connection to rabbit mq
 func ConnectToRabbit(config *RabbitConfig, log *logrus.Entry) (*Consumer, error) {
-	if err := ValidateRabbitConfig(config); err != nil {
+	var err error
+	if err = ValidateRabbitConfig(config); err != nil {
```
rybit: personally, I keep var scoped down in if statements. It stops from accidentally using the wrong `err` b/c of shadowing.

messaging/rabbit.go
```diff
@@ -204,11 +206,20 @@ func ValidateRabbitConfigStruct(servers []string, exchange ExchangeDefinition, q
 
 // ConnectToRabbit will open a TLS connection to rabbit mq
 func ConnectToRabbit(config *RabbitConfig, log *logrus.Entry) (*Consumer, error) {
-	if err := ValidateRabbitConfig(config); err != nil {
+	var err error
```
rybit: declare this in the `if` block you're using it in

messaging/rabbit.go
```diff
@@ -204,11 +206,20 @@ func ValidateRabbitConfigStruct(servers []string, exchange ExchangeDefinition, q
...
+	if err = ValidateRabbitConfig(config); err != nil {
 		return nil, err
 	}
 
-	conn, err := DialToRabbit(config.Servers, config.TLS, log)
+	servers := config.Servers
+	if config.DiscoveryName != "" {
+		servers, err = discoverRabbitServers(config.DiscoveryName)
```
rybit: again, I'd set the `config.Servers` this way when you inspect the object afterwards you don't have a difference from what it is connecting an what it says it is using

messaging/nats.go
sgirones: ah, I like that

messaging/rabbit.go
sgirones: The problem I found is, when I assign the result of discoverRabbitServers to servers and err, if I use :=, it fails because of:
```
servers declared and not used (build)
```
I guess I could have used a tmp var and then assign it to servers?

discovery/main.go
sgirones: Initially I wanted to add more methods to ```Endpoint```, but I guess it is worthless right now

PR #39, opened 2017-10-03
Use async err handler not err returns

rybit:
We don't really care about these at the moment, but we do want a way to get them if they occur

### DIFF
--- a/metrics/transport/signalfx_test.go
+++ b/metrics/transport/signalfx_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/netlify/netlify-commons/metrics"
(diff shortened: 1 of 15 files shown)


metrics/transport/signalfx_test.go
```diff
@@ -15,8 +15,7 @@ func TestWriteToSFX(t *testing.T) {
 	require.NoError(t, err)
 	env := metrics.NewEnvironment(trans)
 	c := env.NewCounter("test.unittest.1", metrics.DimMap{"test": true})
-
-	require.NoError(t, c.Count(nil))
+	c.Count(nil)
```
brycekahle: We should probably add the `assert.Fail` error handler like you have in other tests.

PR #41, opened 2017-10-06
add an optional bugsnag hook

rybit:
take a look at this [bug report](https://app.bugsnag.com/netlify/bitballoon-production/errors/59d7e8a8a27120001878c890?filters%5Bevent.since%5D%5B%5D=1h&filters%5Berror.status%5D%5B%5D=open)

```

import (
	"errors"
	"fmt"
	"os"
	"time"

	"github.com/netlify/netlify-commons/nconf"
)

func main() {
	log, err := nconf.ConfigureLogging(&nconf.LoggingConfig{
		BugSnag: &nconf.BugSnagConfig{APIKey: os.Getenv("BUGSNAG_KEY")},
		Level:   "debug",
	})
	if err != nil {
		panic(err)
	}
	<-time.After(time.Second)
	log.WithError(errors.New("FAIL TEST")).Error("Test")
	<-time.After(time.Second)
	fmt.Println("exiting")
}
```

It looks like there is a bit of a race here ~ we _could_ add a recover block too, but that won't catch if we return/exit too fast.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 339fd13f8046318b0a0df3f78ef85db9fc7536ed1e15fb5223109515f1ed2bf0
-updated: 2017-10-03T10:49:34.76139528+02:00
+hash: c8eb62cdd2b47fd7b2e122f7cdb450721344935bfb5ba449152f9b7fd7442d59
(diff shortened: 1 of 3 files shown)


calavera:
Nice, this is going to be super useful

PR #53, opened 2018-01-26
[mongo] mgo does topology discovery by default. Make it configurable

sgirones:
Make it optional to discover the MongoDB topology. Only connect to the servers provided in the Servers list

Default behaviour is the same as before


sgirones:
Closed. Not really needed right now

PR #54, opened 2018-02-09
create metering log

rybit:
This should let us have jsonl lines in a separate file to parse out what we need.

I decided to use a global style. I don't love that. But it was the best way I could think to not have to pass some logger around *everywhere* or very far down into call chains.

### DIFF
--- a/metering/logging.go
+++ b/metering/logging.go
@@ -11,28 +11,17 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var encoder *json.Encoder
-var meteringBuffer *bufio.Writer
-var logger *logrus.Entry
-var writelock sync.Mutex
-
 func init() {
 	initFromEnv()
 }
 
-func initFromEnv() {
-	fname := os.Getenv("METERING_FILENAME")
-	if fname != "" {
-		err := SetMeteringFile(fname)
-		if err != nil {
-			panic(errors.Wrapf(err, "Failed to open file '%s' which was specified by the env var 'METERING_FILENAME'", fname))
-		}
-	} else {
-		meteringBuffer = bufio.NewWriter(os.Stdout)
-		encoder = json.NewEncoder(meteringBuffer)
-	}
+var global *MeteringLog
+var errorLogger *logrus.Entry = logrus.WithField("component", "metering_errors")
 
-	logger = logrus.WithField("component", "metering")
+type MeteringLog struct {
+	writelock *sync.Mutex
+	encoder   *json.Encoder
+	buffer    *bufio.Writer
 }
 
 type MeteringEvent struct {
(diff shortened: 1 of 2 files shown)


rybit:
also "requiring" some kind of flush was also not my favorite. I'd prefer to get some "the system is shutting down" signal. But I don't know about using just the SIGTERM or SIGINT directly

brycekahle:
Shouldn't the decision of global or not be in each app, instead of the library demanding it?

metering/logging.go
```diff
@@ -0,0 +1,81 @@
...
+		return err
+	}
+
+	meteringBuffer = bufio.NewWriter(open)
+	encoder = json.NewEncoder(meteringBuffer)
+	return nil
+}
+
+func Flush() {
+	writelock.Lock()
```
brycekahle: logrus already has mutexes, so this should not be necessary.

metering/logging.go
```diff
@@ -0,0 +1,81 @@
...
+func initFromEnv() {
+	fname := os.Getenv("METERING_FILENAME")
+	if fname != "" {
+		err := SetMeteringFile(fname)
+		if err != nil {
+			panic(errors.Wrapf(err, "Failed to open file '%s' which was specified by the env var 'METERING_FILENAME'", fname))
+		}
+	} else {
+		meteringBuffer = bufio.NewWriter(os.Stdout)
+		encoder = json.NewEncoder(meteringBuffer)
```
brycekahle: Why aren't we using logrus with a JSONFormatter?

rybit: There is no way to flush logrus I believe.

rybit: I guess I could set logrus to use a buffered output and then flush it manually (like this).

rybit: And logrus has a "msg" field by default - which we don't need

rybit:
you're right on the global part. I am going to make it parrot the metering package

rybit:
@brycekahle updated to make scoping optional

brycekahle:
I was thinking more along the line of a normal constructor function. If the app wishes to make that global, then they store the result in a global variable. There is no need for a global variable in the commons library.

rybit:
I think that this makes it more user-friendly, Are you thinking something like:

```
mLog := metering.NewMeteringLog("path/to/file")
mLog.Record(event, data)
....
mLog.Flush()
```

And it could store the `mLog`?

brycekahle:
@rybit Yes, although `metering.NewLog` would be better (to prevent stutter, see https://golang.org/doc/effective_go.html#package-names). The scope of the object is not the concern of the library.

rybit:
A lot of libraries provide global options for usability. It makes it so you don't _have_ to worry about scope or sharing, that can be handled by the lib. 

Also - yes I am usually good about names & scope but missed one :)

brycekahle:
In general, I'm not a fan of globals. You are hiding a dependency, not getting rid of it.

PR #60, opened 2018-04-25
Queue up metrics to report on interval instead of immediately

brycekahle:
This is currently only implemented for SFX. It also adds a separate config value for the cumulative counter reporting interval (which will actually report to SFX on the next tick of the former interval).

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -43,11 +43,10 @@ func (m *metric) SetTimestamp(t time.Time) {
 }
 
 // AddDimension will add this dimension with locking
-func (m *metric) AddDimension(key string, value interface{}) *metric {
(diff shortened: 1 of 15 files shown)


metrics/metrics.go
```diff
@@ -43,31 +47,30 @@ func (m *metric) SetTimestamp(t time.Time) {
 }
 
 // AddDimension will add this dimension with locking
-func (m *metric) AddDimension(key string, value interface{}) *metric {
+func (m *metric) AddDimension(key string, value interface{}) {
 	m.dimlock.Lock()
 	defer m.dimlock.Unlock()
```
calavera: I guess you can also move this `Unlock` after the assignment and remove the `defer`.

calavera:
Just a minor comment, otherwise LGTM

PR #68, opened 2018-11-12
Support YAML configurations

sgirones:
If the config file has a YAML extension, parse it as a YAML. Otherwise use JSON (keep old behaviour)

Why we want this now? Because it's easier to generate YAML config files using consul-templates, and we are moving netlify-server to K8S, where we use consul-template.


sgirones:
Tested in a test instance and it is able to read and parse a YAML file

PR #72, opened 2019-03-28
remove graceful

rybit:
Since go 1.8 there has been built into the stdlib the ability to graceful using http.Shutdown and .Close

or use the Facebook one


mraerino:
Can you provide a snippet of how one should launch a graceful server that accepts all necessary signals?
Along these lines? https://gist.github.com/peterhellberg/38117e546c217960747aacf689af3dc2#gistcomment-2118438

rybit:
https://gist.github.com/rybit/6bac2de795851df3aaa29139b954175f

rybit:
pretty much on the same lines of the other you linked

mraerino:
Ok, so that should be included in the router package?

rybit:
I think not, each API will have to figure out how to handle its own shutdown. For instance, what else needs to be triggered? Is there an order of operations? What about websockets. I think that starting an http server and closing it yourself should be on the project, not the shared lib

mraerino:
So that graceful boilerplate should be copied into every object?
What is the problem with having the graceful server in a common package?

rybit:
the only part that needs to be 'graceful' is the http server connection. Each service should do it's own cleanup, for instance, not leaving dangling file descriptors, highjacked connections (e.g. websockets or TCP sessions), db connections, timeouts, etc . The only part that is boiler plate is the:
```
httpSvr.Shutdown(....)
```

which is formulaic go

irgendwr:
Hi @rybit , [https://github.com/netlify/git-gateway](netlify/git-gateway) still depends on the now deleted 'graceful' package, so this pull-request broke it (https://github.com/netlify/git-gateway/issues/34).

PR #75, opened 2019-03-28
add tracing framework

rybit:
I found that we were writing the same tracing package in a few places. We also have some that use the chi router and some that use a requestTracker. This should standardize that and make them cross compatible. It doesn't use the chi's keys for the context, but it should still be ok.


tracing/logging.go
```diff
@@ -0,0 +1,73 @@
...
+	entry := getEntry(r)
+	if entry == nil {
+		return logrus.StandardLogger().WithFields(fields)
+	}
+
+	entry.Logger = entry.Logger.WithFields(fields)
+	return entry.Logger
+}
+
+func getLogger(r *http.Request, log logrus.FieldLogger) logrus.FieldLogger {
```
mraerino: This function is never used? Did you intend to use it in `GetLogger`?

mraerino:
@rybit can you rebase this onto master so we can use the newest stuff from there when installing from this branch?

rybit:
@mraerino ~ it is already rebased and pushed? I think you have to run go get again to get any updates

mraerino:
How do I pass the datadog config to this? Is there some autoconfiguration?

@ingride added this to my service: https://github.com/netlify/analytics/blob/master/config/configuration.go#L34 but I'm unsure how to configure the tracing package with it

rybit:
The way that we've been doing it is like this in the setup method:

```
var t opentracing.Tracer = opentracing.NoopTracer{}
	if config.TracingConf.Enabled {
		tracerAddr := fmt.Sprintf("%s:%s", config.TracingConf.Host, config.TracingConf.Port)
		tracerOps := []tracer.StartOption{
			tracer.WithServiceName("netlify-server"),
			tracer.WithAgentAddr(tracerAddr),
		}

		for k, v := range config.TracingConf.Tags {
			tracerOps = append(tracerOps, tracer.WithGlobalTag(k, v))
		}

		t = opentracer.New(tracerOps...)
		log.Infof("Sending traces for to %s", tracerAddr)
	}
	opentracing.SetGlobalTracer(t)
```

rybit:
Also - it seems like I should pull that tracing config into the tracing or nconf package

mraerino:
It would be great to have a `ConfigureTracing` method or something!

mraerino:
@rybit somehow my change is not in this branch? is this the result of your force-push?
maybe try to integrate my branch manually? https://github.com/mraerino/netlify-commons/tree/add-tracing-package-config

rybit:
what is strange is that I did merge that in and couldn't find the code anywhere. I think the parts you want are in the tracing/config.go package

mraerino:
Oh, you are totally right, sorry

sgirones:
I rebased this branch. How far are we from merging? it seems it's being used by analytics and nats-streamer, so would be great to have this on master soon.

rybit:
this is used and ok for merge (doing it now). Any improvements can be done in a new PR

PR #83, opened 2019-05-31
[nats] Implement more auth methods

sgirones:
Implement new auth methods for NATS:
* user/password auth
* token auth

More info: https://nats.io/documentation/server/gnatsd-authentication/

Requirement for: https://github.com/netlify/nats-streamer/pull/2

### DIFF
--- a/messaging/nats.go
+++ b/messaging/nats.go
@@ -43,14 +43,19 @@ func ConfigureNatsConnection(config *nconf.NatsConfig, log logrus.FieldLogger) (
 }
 
 func ConnectToNats(config *nconf.NatsConfig, opts ...nats.Option) (*nats.Conn, error) {
-	if config.TLS != nil {
+	if config.Auth.Method == nconf.NatsAuthMethodUser {
+		opts = append(opts, nats.UserInfo(config.Auth.User, config.Auth.Password))
+	} else if config.Auth.Method == nconf.NatsAuthMethodToken {
+		opts = append(opts, nats.Token(config.Auth.Token))
+	} else if config.Auth.Method == nconf.NatsAuthMethodTLS || config.TLS != nil {
 		tlsConfig, err := config.TLS.TLSConfig()
 		if err != nil {
 			return nil, errors.Wrap(err, "Failed to configure TLS")
 		}
-		if tlsConfig != nil {
-			opts = append(opts, nats.Secure(tlsConfig))
-		}
+
+		opts = append(opts, nats.Secure(tlsConfig))
+	} else {
+		return nil, fmt.Errorf("Invalid auth method: '%s'", config.Auth.Method)
 	}
 
 	return nats.Connect(config.ServerString(), opts...)
--- a/messaging/nats.go
+++ b/messaging/nats.go
@@ -3,6 +3,7 @@ package messaging
 import (
 	"fmt"
 	"io/ioutil"
+	"strings"
 	"time"
 
 	"github.com/nats-io/go-nats"
@@ -24,9 +25,14 @@ func ConfigureNatsConnection(config *nconf.NatsConfig, log logrus.FieldLogger) (
 	if log == nil {
 		log = silent
 	}
+
 	if config == nil {
 		log.Debug("Skipping nats connection because there is no config")
 		return nil, nil
+	}
+
+	if !config.TLS.Enabled {
+		log.Warn("Connection to NATS servers is NOT using TLS")
 	}
 
 	if err := config.LoadServerNames(); err != nil {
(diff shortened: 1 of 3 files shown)


messaging/nats.go
```diff
@@ -43,14 +43,19 @@ func ConfigureNatsConnection(config *nconf.NatsConfig, log logrus.FieldLogger) (
 }
 
 func ConnectToNats(config *nconf.NatsConfig, opts ...nats.Option) (*nats.Conn, error) {
-	if config.TLS != nil {
+	if config.Auth.Method == nconf.NatsAuthMethodUser {
```
rybit: personally, I'd use a switch
```
switch strings.ToLower(config.Auth.Method) {
case "": 
....
}
```

sgirones: The 3rd if is a bit more complex, that's why I didn't use a switch/case.

I refactored the TLS config and now it can use a switch

sgirones:
ready to review again @rybit. I had to split the key/cert and CA loading because with the new NATS auth methods, we need the CA but not the client certificate

messaging/nats.go
```diff
@@ -43,14 +49,28 @@ func ConfigureNatsConnection(config *nconf.NatsConfig, log logrus.FieldLogger) (
...
+		opts = append(opts, nats.Secure(tlsConfig))
+	}
+
+	switch strings.ToLower(config.Auth.Method) {
+	case nconf.NatsAuthMethodUser:
+		opts = append(opts, nats.UserInfo(config.Auth.User, config.Auth.Password))
+	case nconf.NatsAuthMethodToken:
+		opts = append(opts, nats.Token(config.Auth.Token))
+	case nconf.NatsAuthMethodTLS:
+		// if using TLS auth, make sure the client certificate is loaded
```
rybit: I think that doing the TLS stuff in here works well 

```
case nconf.NatsAuthMethodTLS:
  if !config.TLS.Enabled { return an error }
  
  tlsConfig, err := config.TLS.TLSConfig()
  if err != nil {
    return nil, errors.Wrap(err, "Failed to configure TLS config")
  }
  if len(tlsConfig.Certificates) == 0 {
    return nil, fmt.Errorf("TLS auth method is configured but no certificate was loaded")
  }
  
  opts = append(opts, nats.Secure(tlsConfig))
default: 
...
```

nconf/tls.go
```diff
@@ -21,88 +24,89 @@ type TLSConfig struct {
...
+	tlsConf := &tls.Config{
+		MinVersion:         tls.VersionTLS12,
+		InsecureSkipVerify: cfg.Insecure,
 	}
 
-	if tlsconf != nil {
-		tlsconf.InsecureSkipVerify = cfg.Insecure
+	// Load CA
+	if cfg.CA != "" {
+		logrus.Debugf("Loading CA value")
```
rybit: i wouldn't add a statement here

nconf/tls.go
```diff
@@ -21,88 +24,89 @@ type TLSConfig struct {
...
-		}
-		pool = p
-	} else {
-		pool = x509.NewCertPool()
-		if !pool.AppendCertsFromPEM([]byte(ca)) {
-			return nil, fmt.Errorf("Failed to add CA cert")
-		}
+	// Load Certs if any
+	var cert tls.Certificate
+	err = nil
```
rybit: This shouldn't be needed it - you return if it was an error before

nconf/tls.go
rybit: Generally I don't like adding built in log statements to a library

messaging/nats.go
sgirones: we want the TLS conf always, even if the auth is not TLS, because we still need encrypted connections to the servers.

The only difference is for `tls` auth, we also need the certificate. For the rest, we only need the RootCAs

nconf/tls.go
sgirones: fair. I wanted to know if the CA/Certificate was loaded and from where. If NATS client fails to connect because the CA is missing, it won't hint you in that direction and it took me some testing and debugging to figure out that the env var was incorrect and the CA was not loaded

nconf/tls.go
sgirones: argh, of course. Forgot to delete this

messaging/nats.go
rybit: ah good point I didn't see that. If that is the case then I'd probably want to be more careful around:
```
if len(tlsConfig.Certificates) == 0
```

`tlsConfig` could be nil I think (shouldn't be) but it would cause a panic instead of an error message

sgirones:
Refs https://github.com/netlify/ansible/issues/2649

sgirones:
@rybit can you do a final review?

sgirones:
I verified this code also works with cachecontroller

PR #84, opened 2019-07-23
Add a generic router interface

rybit:
This is based on the stuff from the analytics just stripped down a bit. Also not using a test suite.


router/router.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/sebest/xff"
+	"github.com/sirupsen/logrus"
+)
+
+type chiWrapper struct {
+	chi chi.Router
+}
+
+// Options configures specifics of the default router
+type Options struct {
```
smoya: I'm more in favor of functional options as a configuration mechanism: 
* https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis
* https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html

In a first look, it could seem weird but it facilitates final users to compose semantically the options to pass, becoming more scalable from the point of view of the code as easy to test and distribute. 

Some big Golang public libs use this pattern:
* New preview version of AWS sdk: https://github.com/aws/aws-sdk-go-v2/blob/a3571312476e84b6837ca42f6c8a4654633d0a53/aws/request.go#L142
*https://github.com/kubernetes/kubernetes/blob/a33840e0232aaf2b5c14c1c754d9189d7e841ec1/pkg/kubelet/container/runtime.go#L611
* https://github.com/uber/kraken/blob/128c7e2e4c264c780edc2d764c6ec12484e1dac8/lib/backend/gcsbackend/client.go#L75
* https://github.com/gorilla/csrf/blob/master/options.go#L8

rybit:
ok @smoya I'm taking [this](https://github.com/netlify/netlify-commons/pull/84#discussion_r313410228) out of the thread so that we don't lose it in pushes. re:options as functions I pushed up a change and generally I like that but one thing that I'm struggling with is that these make it hard to have an order between options. 
For instance, these two things have different behavior
```
// will have the version header on the health endpoint
New(log, OptVersionHeader("a", "b"), OptHealthCheck("/health"))
```

```
// will NOT have the version header on the health endpoint
New(log, OptHealthCheck("/health"),OptVersionHeader("a", "b"))
```

Because these ultimately funnel into a middleware stack, which the consumer shouldn't be aware of.

rybit:
trying a slightly different approach

router/router.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	r.Use(VersionHeader(r.svcName, r.version))
+	if r.enableCORS {
+		corsMiddleware := cors.New(cors.Options{
+			AllowedMethods:   []string{"GET", "POST", "PATCH", "PUT", "DELETE"},
+			AllowedHeaders:   []string{"Accept", "Authorization", "Content-Type"},
+			ExposedHeaders:   []string{"Link", "X-Total-Count"},
+			AllowCredentials: true,
+		})
+		r.Use(corsMiddleware.Handler)
```
smoya: Should we also add `OPTIONS` route and handler?

smoya:
@rybit After reviewing your latest commits and considering the nature of this piece of software, you made me realize that probably functional options are not a fit for it, due to the fact that order of middlewares matters. However I think that this pattern would apply for most of configurable package services we have.

IMHO I would revert the latest commits and leave as it was prior to my intrusion!

router/middleware.go
```diff
@@ -0,0 +1,95 @@
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			f(w, r, next)
+		})
+	}
+}
+
+func VersionHeader(serviceName, version string) Middleware {
+	return MiddlewareFunc(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
+		next.ServeHTTP(w, r)
```
smoya: Is this made in purporse?: We call the whole middleware chain and right after we add the version header?

rybit: yeah that is my intent, we are setting the version header for the client, not our internal consumption. This makes it so that none of our code can override that header. 

Internal code can get access to the `Version` directly

rybit:
ok @smoya we talked a bit about this but I wanted a paper trail. I don't like this last one either but, I'd like to ship with my latest impl because of consistency. It is kinda ugly, but not too bad. I think moving us to a functional configuration pattern is good. 

I'm going to merge this on Monday my time unless you say no (which I'm totally ok with b/c I'm waffling pretty hard on this choice).

smoya:
As I mentioned, functional opts are not adding value to this code. However, the fact that we decided to give a try to them using it from now on in new developments is completely aligned with your suggestion of using them for the `router` in order to keep consistency.

Happy with the conclusion. Happy to have this merged 👍

rybit:
igm

PR #85, opened 2019-08-13
Import router package from analytics repository

smoya:
Moving the `router` package from https://github.com/netlify/analytics/tree/master/router to `netlify-commons` in order to make it available for any other Golang repository.


smoya:
Closed as duplicated of https://github.com/netlify/netlify-commons/pull/84

PR #87, opened 2019-08-16
update nats version and cleanup config

rybit:
We were using an old library and the project has moved. I also cleaned up some of the config.


rybit:
Ok after talking with @smoya we decided that we'd go with this scheme: 
- config in the package is the default
- if you can't (import loops, etc) then we move the _whole project_ to use the central config package. 

The goal is to have a consistent approach across the project. Not some packages with their own config and some in the `conf` package

PR #88, opened 2019-08-21
Mount router health check as route instead of using a middleware

smoya:
After finding a bug in the `HealthCheck` middleware (See comment: TBD), I realise that simulating and endpoint via middleware instead of mounting it is not giving us any value. .

It makes more sense to directly mount the health check in the router since at the end it does not manipulates the request chain or triggers a side effect, being just the latest handler for its path.

### DIFF
--- a/router/middleware.go
+++ b/router/middleware.go
@@ -79,17 +79,3 @@ func Recoverer(w http.ResponseWriter, r *http.Request, next http.Handler) {
 
 	next.ServeHTTP(w, r)
 }
-
-func HealthCheck(route string, f APIHandler) Middleware {
-	return MiddlewareFunc(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
-		if r.URL.Path == route {
-			if f == nil {
-				w.WriteHeader(http.StatusOK)
-				return
-			}
-			HandleError(f(w, r), w, r)
-			return
-		}
-		next.ServeHTTP(w, r)
-	})
-}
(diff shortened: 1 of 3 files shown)


router/middleware.go
```diff
@@ -79,17 +79,3 @@ func Recoverer(w http.ResponseWriter, r *http.Request, next http.Handler) {
...
 }
-
-func HealthCheck(route string, f APIHandler) Middleware {
-	return MiddlewareFunc(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
-		if r.URL.Path == route {
-			if f == nil {
-				w.WriteHeader(http.StatusOK)
-				return
-			}
-			HandleError(f(w, r), w, r)
```
smoya: If the error is `nil`, the HandleError function will fail in this line https://github.com/netlify/netlify-commons/blob/master/router/errors.go#L92. This is one of the pain points of nil interfaces.

router/router.go
```diff
@@ -71,7 +71,12 @@ func New(log logrus.FieldLogger, options ...Option) Router {
 	}
 
 	if r.healthEndpoint != "" {
-		r.Use(HealthCheck(r.healthEndpoint, r.healthHandler))
+		h := r.healthHandler
```
rybit: I think that there is an issue with adding a route and then adding middlewares. Which also means that even if we move the tracing one up (which would suck) ~ it stops users of the lib from adding more middlewares.

smoya: You're completely right: 
>panic: chi:  all middlewares must be defined before routes on a mux

I'm gonna drop this PR and fix the bug on the Middleware then!
Thanks for the catch!

PR #98, opened 2019-11-05
Build static godoc

mraerino:


calavera:
we should merge this and write a blog post @mraerino :wink:

PR #102, opened 2019-12-17
Expose Mount from the chi Router.

calavera:
So you can mount other custom handlers.


calavera:
> we've often done

I don't want to enforce the HTTPError that our custom handler enforces, because of reasons :wink:

calavera:
It's also nicer to actually have access to the real router, and not a trimmed version.

PR #105, opened 2020-01-10
Bugsnag updates

mheffner:
Looking to test this out on eventeer, so reviving the original implementation here. Two updates:

1. Passed the current app version to BugSnag config in order to track releases in the app.
2. Updates to the latest golang release.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/bitly/go-simplejson v0.5.0 // indirect
 	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
-	github.com/bugsnag/bugsnag-go v1.5.1
+	github.com/bugsnag/bugsnag-go v1.5.3
 	github.com/bugsnag/panicwrap v1.2.0 // indirect
 	github.com/globalsign/mgo v0.0.0-20181015135952-eeefdecb41b8
 	github.com/go-chi/chi v4.0.2+incompatible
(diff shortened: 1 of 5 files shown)


rybit:
the changes look good - I'd try playing with it in eventeer without merging and see if it adds value. I'm not sure that it being in the hook will work well

mheffner:
> the changes look good - I'd try playing with it in eventeer without merging and see if it adds value. I'm not sure that it being in the hook will work well

Sounds good, I'll give that a shot. We could also make the hook behavior opt-in with a config option I guess.

rybit:
At the time of writing the component I was of the mindset “configure it and it is on” now I am more of the “explicit config option” mindset. 

Lmk if this work well.

mheffner:
Testing this out in https://github.com/netlify/eventeer/pull/31

I updated this PR to pull Bugsnag configuration out of the logging configuration and make the log hook integration optional. Since we are making it optional, it didn't seem appropriate to configure it with the logging config.

mraerino:
Would this mean that every service that is using bugsnag now would need to change their config structure when upgrading netlify-commons?

mheffner:
> Would this mean that every service that is using bugsnag now would need to change their config structure when upgrading netlify-commons?

It would yes, do you have an idea of how many services are using Bugsnag right now?

rybit:
to my knowledge - 0 go services use bugsnag atm

mheffner:
@rybit You ok with this change? Seems to work in eventeer so far, though log hook is still opt-in.

rybit:
yeah I'm cool with breaking it out

PR #108, opened 2020-02-03
API lifecycle support

mheffner:
This tweaks the `APIDefinition` interface to include full lifecycle support (Start + Stop). This allows the API to perform a clean shutdown like releasing and removing Consul locks. This was the remaining piece to allow Eventeer to remove its Server implementation.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -85,7 +85,11 @@ func (s *Server) ListenAndServe() error {
 	}
 
 	// Now that server is no longer listening, shutdown the API
+	s.log.Info("Listener shutdown, stopping API")
+
 	s.api.Stop()
+
+	s.log.Debug("Completed shutting down the underlying API")
 
 	if err == http.ErrServerClosed {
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -77,20 +83,36 @@ func (s *Server) ListenAndServe() error {
 	} else {
 		err = s.svr.ListenAndServe()
 	}
+
+	// Now that server is no longer listening, shutdown the API
```
rybit:
```suggestion
	// Now that server is no longer listening, shutdown the API
	s.log.Info("Listener shutdown, stoping API")
```

server/server.go
```diff
@@ -77,20 +83,36 @@ func (s *Server) ListenAndServe() error {
...
+	s.api.Stop()
+
```
rybit:
```suggestion
s.log.Debug("Completed shutting down the underlying API")
```

PR #112, opened 2020-03-26
fix(router): early return when zero vallue of err in HandleError

smoya:
## Issue

Router component wasn't able to handle nil pointer to type `HTTPError`, meaning the following simple handler 

```go
func(_ http.ResponseWriter, _ *http.Request) *HTTPError {
	return nil
}
```

was ending with a panic:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
```

## Reason

`return nil` is returning a nil pointer to type `HTTPError`, which is later on passed to https://github.com/netlify/netlify-commons/blob/91f500e0a2b59a8667d48df8f2cbeea2160da991/router/errors.go#L88

As you can see, the first arguments expects the interface `error`, which is completely fulfilled.
According to [Go FAQS](https://golang.org/doc/faq#nil_error):
> An interface value is nil only if the V and T are both unset

In that case, value is `nil` by type is `*HTTPError` so when this check gave a `false`.
https://github.com/netlify/netlify-commons/blob/eb929b04d4fa127aeb87ecb2ba7d54bad1009049/router/errors.go#L88-L90

## Fix

The ideal fix is checking for `== nil` before calling `HandleError` func as I did here: https://github.com/netlify/netlify-commons/blob/f082f63c611840b9db35a13070e774bd591421c2/router/middleware.go#L94-L96
The main issue is that we could easily fall into doing the same mistake in the future. That's the reason I decided to add an extra check in `HandleError`. The first idea was to use `reflect.ValueOf(err).IsNil()` but then, concerned about performance, I found [this article](https://dev.to/pauljlucas/go-tcha-when-nil--nil-hic) where they suggest to use the `unsafe` package instead with https://github.com/netlify/netlify-commons/blob/f082f63c611840b9db35a13070e774bd591421c2/router/errors.go#L90

### DIFF
--- a/router/errors.go
+++ b/router/errors.go
@@ -3,6 +3,7 @@ package router
 import (
 	"fmt"
 	"net/http"
+	"unsafe"
 
 	"github.com/netlify/netlify-commons/tracing"
 )
@@ -85,7 +86,8 @@ func httpError(code int, fmtString string, args ...interface{}) *HTTPError {
 
 // HandleError will handle an error
 func HandleError(err error, w http.ResponseWriter, r *http.Request) {
-	if err == nil {
+	// if err is nil or nil pointer of type *HTTPError
+	if err == nil || (*[2]uintptr)(unsafe.Pointer(&err))[1] == 0 {
 		return
 	}
 
--- a/router/errors_test.go
+++ b/router/errors_test.go
@@ -3,14 +3,15 @@ package router
 import (
 	"errors"
 	"fmt"
-	"github.com/netlify/netlify-commons/tracing"
-	"github.com/sirupsen/logrus/hooks/test"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"testing"
+
+	"github.com/netlify/netlify-commons/tracing"
+	"github.com/sirupsen/logrus/hooks/test"
+	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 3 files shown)


router/errors.go
```diff
@@ -85,7 +86,8 @@ func httpError(code int, fmtString string, args ...interface{}) *HTTPError {
 
 // HandleError will handle an error
 func HandleError(err error, w http.ResponseWriter, r *http.Request) {
-	if err == nil {
+	// if err is nil or nil pointer of type *HTTPError
+	if err == nil || (*[2]uintptr)(unsafe.Pointer(&err))[1] == 0 {
```
phenixrizen: if err.(*HTTPError) == nil || err == nil will also work and then you don't have to use unsafe

phenixrizen: This introduces unsafe into the netlify-commons package which is used by most services. therefore introducing this in them as well.  I checked the code and it is valid but there is a better way of doing this with type assertion.  You can also run go vet to check if we are doing something not so smart with unsafe.  Unsafe steps around Go's type safety and in my opinion should only be used way no other way is possible

rybit: Agree on not using `unsafe` if we can avoid it

smoya: The issue is not about using `unsafe`. The issue is the function signature and the responsibilities of it.  The function could receive **any** implementation of `error` that ends up with the same problem.

Btw, I recommend you to run a grep on all dependencies looking for the usage of `unsafe`. 
Feel free to open a new PR with the needed changes if you agree.

PR #113, opened 2020-03-26
removed unsafe from library

phenixrizen:


phenixrizen:
@mheffner I reverted the mod changes,  I'm so use to running go mod tidy before I commit.

PR #114, opened 2020-04-01
Pull in AtomicBool from eventeer.

mheffner:
Simple wrapper around Int32 atomic's to provide an atomic boolean.

### DIFF
--- a/util/atomic_bool.go
+++ b/util/atomic_bool.go
@@ -23,17 +23,14 @@ func (a *AtomicBool) Set(boolValue bool) bool {
 	if boolValue {
 		intValue = trueValue
 	}
-	return a.toTruthy(atomic.SwapInt32(&a.value, intValue))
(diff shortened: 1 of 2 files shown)


util/atomic_bool.go
```diff
@@ -0,0 +1,39 @@
...
+// Get will return the current value
+func (a *AtomicBool) Get() bool {
+	return a.toTruthy(atomic.LoadInt32(&a.value))
+}
+
+func (a *AtomicBool) toTruthy(val int32) bool {
+	if val == falseValue {
+		return false
+	}
+	return true
```
rybit, line 38: minor thing but you can do `return val == trueValue`

rybit: also this doesn't need to be on the `AtomicBool` as a method.

mheffner: sounds good, changed slightly to test against false but still single line

PR #115, opened 2020-04-02
Slight interface tweak for AtomicBool

mheffner:
This is a slight tweak to the previous AtomicBool interface. Essentially as a consumer instead of:

```
struct {
isRunning *util.AtomicBool
...
}
```

you use:
```
struct {
isRunning util.AtomicBool
...
}
```

This hides the pointer abstraction behind an interface for the consumer.

### DIFF
--- a/util/atomic_bool.go
+++ b/util/atomic_bool.go
@@ -7,18 +7,23 @@ const (
 	trueValue  int32 = 1
 )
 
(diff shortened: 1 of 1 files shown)


mheffner:
Welp, I had thought about making this change when I pulled it in from Eventeer but forgot, cc @rybit

PR #117, opened 2020-04-07
Import metriks mod from Eventeer.

mheffner:
Bring over the metriks wrapper from Eventeer so that we can incorporate this in other projects, like buildbot. This is mostly a nicer wrapper around the [go-metrics](https://github.com/armon/go-metrics) library, which adds things like automatic [runtime metrics](https://github.com/armon/go-metrics/blob/master/metrics.go#L244-L250). These wrappers make it simple to incorporate statsd metrics in projects with very little instrumentation overhead.

This import also includes a `DBStats` instrumentation that will, if used, will automatically instrument the connection runtime stats of a DB connection/pool.

This mod is named `metriks` to specifically avoid collisions with the underlying "metrics" module. This isn't specifically required but it did seem to help clean up the abstraction a bit.


mheffner:
Looks like Go 1.12 is broken so I need to fix

metriks/metriks.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"github.com/armon/go-metrics"
+	"github.com/armon/go-metrics/datadog"
+)
+
+const timerGranularity = time.Millisecond
+
+// Init will initialize the internal metrics system and build a Datadog statsd sink
+func Init(serviceName string, conf Config) error {
+	sink, err := datadog.NewDogStatsdSink(statsdAddr(conf), conf.Name)
```
smoya, line 15: I don't think Datadog should be the default (and the only one) sink. We may want to initialize an inmemory sink (for testing) or a discard all for environments where we don't care about metrics (development).

metriks/metriks.go
```diff
@@ -0,0 +1,124 @@
...
+	for k, v := range conf.Tags {
+		tags = append(tags, fmt.Sprintf("%s:%s", k, v))
+	}
+
+	sink.SetTags(tags)
+
+	c := metrics.DefaultConfig(serviceName)
+	c.TimerGranularity = timerGranularity
+
+	if _, err := metrics.NewGlobal(c, sink); err != nil {
```
smoya, line 30: Setting a global metrics object won't allow us to eventually use different sinks in the same app. I.e., we may want to use Datadog for one mod but Prometheus for another one. I want also to emphasize the fact that this could cause problems in tests around metrics and collisions between tests in case different sinks need to be used.

metriks/db_stats.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (d *dBStats) Stop() {
+	d.sched.Stop()
+}
+
+func (d *dBStats) emitStats() {
+	stats := d.db.Stats()
+
+	d.emitStat("MaxOpenConnections", float32(stats.MaxOpenConnections))
```
smoya, line 54: What if we want to add more metrics or in the opposite, don't send some?

metriks/metriks.go
```diff
@@ -0,0 +1,124 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/armon/go-metrics"
+	"github.com/armon/go-metrics/datadog"
+)
+
+const timerGranularity = time.Millisecond
```
smoya, line 11: What about making this configurable? Don't see a clear value of having this const here.

metriks/metriks.go
mheffner: So metrics are one of those cross-cutting concerns that I feel you really want to be able to use anywhere in a code base easily, similar to logging. I think there's a lot of value in keeping support for a global metriks scope, because it means you can immediately instrument any piece of code, no matter how deep in an abstraction layer. From past experience, I've found that if you add barriers to instrumentation, even minor, it can tend to lead to skipping it in places you really would prefer to have it. You don't want to have to balance between refactoring multiple layers of abstraction to pass a metrics object so that you can quickly instrument an error count, for example.

However, the global model works best when it's simply internal service / application code and we've all agreed on a shared instrumentation library. If we were building shared libraries that we wanted to open source and allow others to use in applications outside Netlify, then global metrics registry approach wouldn't work and we'd want a configurable metrics dependency. For this, I think we could always take this current interface and make it a proper exported interface that any open source library could take as an interface and invoke directly on. We probably want to wait until we've settled on one across projects though.

> we may want to use Datadog for one mod but Prometheus for another one

I'm hoping we never want to split metrics for a service tier between different metrics systems. :-D That being said, if we need to we could build an interface from this that certain modules can take a metrics object, instead of using the global. I think that'd be best for code shared outside of our org.

aarushik93:
As part of this pkg, it would be handy if we had a SetTags function, so we can override or configure global tags, following the same abstraction pattern currently being used for the actual metrics function

metriks/metriks.go
rybit: I agree on the global settings. I think that it can be very nice to not have to pass the metrics connection all the way down. It is nice in logs b/c we're able to maintain state (e.g. the request id) but that isn't really needed in the case of metrics. 

> we may want to use Datadog for one mod but Prometheus for another one

No please 😭. I think that the `InitWithSink` is a good idea, we can test with feature flags for percentages of the service, not running two sinks in the same service. Also, we _just_ signed a contract with DataDog, this will be a moot point for at least another year.

mheffner:
I expanded the `Init` method flexibility, allowing a bring-your-own-sink and extra tags.

metriks/metriks.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	sink.SetTags(tags)
+
+	return InitWithSink(serviceName, sink)
+}
+
+// InitWithSink initializes the internal metrics system with custom sink
+func InitWithSink(serviceName string, sink metrics.MetricSink) error {
+	c := metrics.DefaultConfig(serviceName)
```
rybit, line 43: I'm not sure the default config is what we want. 

https://github.com/armon/go-metrics/blob/58588f401c2cc130a7308a52ca3bc6c0a76db04b/start.go#L52-L68

- I'm not sure if the `servicename` is going to end up as a prefix. I am not seeing it set `EnableServiceLabel`. 
- The `EnableHostname` says that it will prefix, which would cause a lot of trouble in the k8s world. I'd disable it.  
- The profiling stuff I wouldn't use, this one I'm more on the fence for this one.

mheffner: We are prefixing metric names by the `serviceName`, this is sort of nice because DDOG makes it difficult to search for metrics by tag in the dashboard UI. So you can type "eventeer" and find all eventeer metrics.

It's a good question on the hostname, I'll need to double check. We are definitely *not* prefixing metric names with the hostname and instead adding a `host` label. However, it does seem like that shouldn't be the default behavior if I'm reading this right so we probably want to be explicit. Either way, in our current configs we are adding the pod name as a label which is more useful than the hostname (which is already added by the ddog agent). So I'd vote to remove the host entirely by default here.

I find the profiling useful for general performance data. The number of goroutines can often grow large in unexpected scenarios and be a source of problems, I like to have it on a dashboard somewhere just in case.

smoya: Regarding the hostname prefix, indeed we *should* not prefix as the host is a dimension per se.
In Edge-State we did the following: 

```
c := gometrics.DefaultConfig(svcName)
	c.TimerGranularity = timerGranularity
	c.HostName = hostname

	c.EnableServiceLabel = false // Otherwise it will add the service as label instead of prefix of metric name
	c.EnableHostname = false     // Otherwise it adds the hostname to all metrics
```

rybit: I think hostname doesn't make sense by default. Same with pod name. In the k8s world I think this could cause a large growth in our cardinality (b/c pods are ephemeral). I think on the edge it could be really valuable to have the hostname (why is this node failing vs this one passing). 

One idea (totally not a requirement) is that this would be to implement a config mgmt using a options style (`func Config(serviceName string, opts ...Option)`).

PR #119, opened 2020-05-01
Keep up with the latest

mheffner:


rybit:
we should move to github actions I think, but not a blocker for this PR

PR #124, opened 2020-05-15
fix - add option to pass in custom router

emilyzhang:


emilyzhang:
jk, this is irrelevant now - we thought some middleware issues in support-api were being caused by the way the router is set up in netlify-commonts, but we were mistaken about the cause of the problems.

PR #131, opened 2020-06-15
take variadic args for labels

rybit:
I noticed in some of our PRs there was some cumbersome code (imho). Essentially you had to do a bunch of boilerplate to add a label. I thought that this would make things cleaner. I am pretty sure that we could pass through to always with labels (or an empty), but this seemed safer

### DIFF
--- a/metriks/metriks_test.go
+++ b/metriks/metriks_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/armon/go-metrics"
 	"github.com/armon/go-metrics/datadog"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/net/nettest"
 )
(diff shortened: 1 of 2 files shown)


metriks/metriks_test.go
```diff
@@ -69,3 +70,20 @@ func TestInMemorySink(t *testing.T) {
 	require.Equal(t, "test.test_counter", met[0].Counters["test.test_counter"].Name)
 	require.Equal(t, 1, met[0].Counters["test.test_counter"].Count)
 }
+
+func TestIncWithLabels(t *testing.T) {
+	sink, err := InitWithURL("test", "inmem://?interval=1s&retain=2s")
+	require.NoError(t, err)
+	require.IsType(t, &metrics.InmemSink{}, sink)
+
+	Inc("test_counter", 1, L("tag", "value"))
```
emilyzhang, line 79: maybe we should do a test with multiple labels as well? up to you though

rybit:
good point @mheffner, I'm torn on removing the old methods so I made them pass throughs

mheffner:
@rybit I guess one question though, for cases where a user might maintain a list of labels through their app, how do you expand that to variadic args in go?

current version:
```
labels := Lables(L("foo", "bar"), L("blah", "tada"))
...
IncLabels("success", labels, 12)
```

rybit:
you can use `...` to expand a slice into a variadic args.

See https://github.com/netlify/netlify-commons/pull/131/files#diff-d3bafce4f5b81d6ea35b0ba56d410486R144-R146

You example would look like this:
```
labels := Lables(L("foo", "bar"), L("blah", "tada"))
...
Inc("success", 12, labels...)
```

PR #133, opened 2020-06-19
Add pprof lib

phenixrizen:
## Summary
This PR adds a pprof lib. This will allow us to debug and profile our services. 
Runs a pprof server on the specified port or `Run()` listens on `localhost:6060`.

Adds the following routes:
- /debug/pprof/
- /debug/pprof/cmdline
- /debug/pprof/profile
- /debug/pprof/symbol
- /debug/pprof/trace

Use in Kubernetes:
```
$ kubectl port-forward pods/netlify-whatever 6060:6060
$ go tool pprof http://localhost:6060/debug/pprof/heap
```

## Test plan
Tests are provided.


pprof/pprof.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"net/http"
+	"net/http/pprof"
+
+	log "github.com/sirupsen/logrus"
+)
+
+var ListenAddress = "localhost:6060"
+
+// This init is needed to disable the default handlers registered during the init() from importing net/http/pprof
```
rybit, line 12: could this have any implications about any other packages using the default mux? Like if they're using it (can't think of anything) we'd be overriding it and they'd not be able to make any matching paths?

Not a blocker or something we have to fix, just a note

phenixrizen: Nope it's scoped per that import of `http` in the `pprof` package.

phenixrizen:
```go
package main

import (
	"io"
	"log"
	"net/http"

	"github.com/netlify/netlify-commons/pprof"
)

func main() {
	h1 := func(w http.ResponseWriter, _ *http.Request) {
		io.WriteString(w, "Hello!\n")
	}
	h2 := func(w http.ResponseWriter, _ *http.Request) {
		io.WriteString(w, "Bar!\n")
	}

	http.HandleFunc("/", h1)
	http.HandleFunc("/foo", h2)

	pprof.Run()

	log.Fatal(http.ListenAndServe(":8888", nil))
}
```
You can run this and see the mux that is created with the HandleFunc's is still intact

rybit: That’s awesome. I didn’t know go would do that.

PR #141, opened 2020-08-06
change the APIHandler add the Recoverer

rybit:
* Adds the option (and defaults on) a recoverer middleware
* changes the `APIHandler` to use `error`

The latter is because of learnings from: https://golang.org/doc/faq#nil_error (thanks @smoya!). It should be transparent to code using it. If using the `HandleError` it already handles this case

If this gets merged I think we should up the version number to indicate a breaking change.

### DIFF
--- a/router/middleware.go
+++ b/router/middleware.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/netlify/netlify-commons/tracing"
+	"github.com/sirupsen/logrus"
 )
 
 var bearerRegexp = regexp.MustCompile(`^(?:B|b)earer (\S+$)`)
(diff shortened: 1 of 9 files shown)


router/middleware.go
```diff
@@ -26,9 +29,12 @@ func MiddlewareFunc(f func(w http.ResponseWriter, r *http.Request, next http.Han
 }
 
 func VersionHeader(serviceName, version string) Middleware {
+	if version == "" {
+		version = "unknown"
+	}
 	return MiddlewareFunc(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
-		next.ServeHTTP(w, r)
 		w.Header().Set(fmt.Sprintf(versionHeaderTempl, strings.ToUpper(serviceName)), version)
+		next.ServeHTTP(w, r)
```
phenixrizen, line 37: Good catch on this!

PR #142, opened 2020-08-06
make tracing and metrics part of the default config

rybit:
The goal here is to have a standard names and configuration that people can use. We can change the boilerplate templates used to provide these by default too.

This is slightly breaking change in that you have to provide the "serviceName" to the args now. We should consider unifying the other calls (e.g. in `server`) too.

### DIFF
--- a/nconf/args_test.go
+++ b/nconf/args_test.go
@@ -37,7 +37,7 @@ PF_LOG_QUOTE_EMPTY_FIELDS=true
 		EnvFile: tmp.Name(),
 	}
 
-	log, err := args.Setup(cfg, "")
+	log, err := args.Setup(cfg, "", "")
 	require.NoError(t, err)
 
 	// check that we did call configure the logger
--- a/metriks/config.go
+++ b/metriks/config.go
@@ -1,8 +1,9 @@
 package metriks
 
 type Config struct {
-	Host string `default:"localhost"`
-	Port int    `default:"8125"`
+	Enabled bool
+	Host    string `default:"localhost"`
+	Port    int    `default:"8125"`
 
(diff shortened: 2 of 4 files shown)


nconf/args_test.go
```diff
@@ -37,7 +37,7 @@ PF_LOG_QUOTE_EMPTY_FIELDS=true
 		EnvFile: tmp.Name(),
 	}
 
-	log, err := args.Setup(cfg, "")
+	log, err := args.Setup(cfg, "", "")
```
emilyzhang, line 40: maybe we should also add tests to check whether the global tracer and metrics have been configured successfully

rybit: I was looking into that a bit ~ It isn't actually easy to access those in their packages. For instance, there is no way to get the `Sink` in the metriks package. I didn't think it worth the effort personally.

vbrown608:
We were bitten by this in buildbot, as `metriks` is now disabled by default. Maybe I missed an announcement, if not one would be very helpful in the future for breaking changes to go-commons, or a backwards-compatible zero value like `Disabled` :pray:

mheffner:
Looks like this also caused netlify-server metrics to stop reporting several days ago: https://app.datadoghq.com/dashboard/p8i-t9f-3vt/netlify-server-timeboard?from_ts=1597248727658&fullscreen_section=overview&fullscreen_widget=4277456194287042&live=true&to_ts=1597853527658&tpl_var_env=production&fullscreen_start_ts=1597248777765&fullscreen_end_ts=1597853577765&fullscreen_paused=false. Agreed we should be careful about changing the defaults without some explicit change process.

PR #143, opened 2020-08-06
make the request ID accessible from context

rybit:
In cert-store we were trying to pull the request ID from the context, so we had to jam it in there. This is probably a common pattern (e.g. going to make a secondary request).

I decided to use the `Tracer` (vs adding an explicit key) so that there was no chance of drift. Plus it made it easy. I'm def open if people don't like this (I kinda don't like this whole package).

### DIFF
--- a/tracing/logging.go
+++ b/tracing/logging.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/http"
 
+	uuid "github.com/satori/go.uuid"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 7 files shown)


tracing/logging.go
```diff
@@ -11,7 +11,7 @@ const (
 	logKey = contextKey("nf-log-key")
 )
 
-func requestLogger(r *http.Request, log logrus.FieldLogger) logrus.FieldLogger {
+func requestLogger(r *http.Request, log logrus.FieldLogger) (logrus.FieldLogger, string) {
```
rybit, line 14: this change is mostly to not call the `RequestID` multiple times. Only used in 1 place (and not exported) so I tossed it in

PR #144, opened 2020-08-11
Update mongo driver

mheffner:
Update the mgo driver to the official Mongo go driver. I changed the pkg name slightly to avoid the name collision.

We are using this in several projects in production now, this will officially move it into master.


mongoclient/db.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+type Config struct {
+	AppName            string
+	TLS                *nconf.TLSConfig
+	Servers            []string
+	ReplSetName        string
+	ConnTimeout        time.Duration
+	Auth               *Auth
+	SecondaryPreferred bool
```
rybit, line 38: If you're not specifying `split_words:"true"` then this is going to be kinda hackey in the env files.

mongoclient/db.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Connect connects to MongoDB
+func Connect(config *Config, log *logrus.Entry) (*mongo.Client, error) {
```
rybit, line 42: What do you think about using a `Option` driven API (like the `router` package).

Something like 
```
type Option func(opt *options.ClientOptions) error

func Connect(log logrus.FieldLogger, replSet string, servers []string, opts ...Option) (*mongo.Client, error) {
	finalOpts := options.Client().
		SetConnectTimeout(time.Minute). // use a default
		SetReplicaSet(replSet).
		SetHosts(servers)
	for _, opt := range opts {
		if err := opt(finalOpts); err != nil {
			return nil, err
		}
	}
	log.WithFields(logrus.Fields{
		"servers":     strings.Join(finalOpts.Hosts, ","),
		"replica_set": finalOpts.ReplicaSet,
	}).Debug("Dialing database")

	client, err := mongo.Connect(context.Background(), finalOpts)
	if err != nil {
		return nil, err
	}

	// Connect does not block for server discovery, so we should ping to ensure
	// we are actually connected
	ctx, cancel := context.WithTimeout(context.Background(), *finalOpts.ConnectTimeout+time.Second)
	defer cancel()

	if err := client.Ping(ctx, readpref.Primary()); err != nil {
		log.WithError(err).Error("Failed to ping primary Mongo instance")
		return nil, err
	}

	log.Debug("Connected to MongoDB")
	return client, nil
}

func TLSOption(config *nconf.TLSConfig) Option {
	return func(opts *options.ClientOptions) error {
		tlsConfig, err := config.TLSConfig()
		if err != nil {
			return err
		}
		opts.SetTLSConfig(tlsConfig)
		return nil
	}
}

func AuthOption(config Auth) Option {
	return func(opts *options.ClientOptions) error {
		creds := options.Credential{
			Username:   config.Username,
			Password:   config.Password,
			AuthSource: config.Source,
		}

		opts.SetAuth(creds)
		return nil
	}
}
```

It would expose the future direct mongo options too so that we don't have to tweak the API a lot. 

One drawback is that you can't directly import `Config` struct and just start using it. (though we could have 2 methods `ConfigWithOptions` and `Config`)

mongoclient/db.go
mheffner: Yeah, so I intentionally didn't add the config file tags here because I wanted to separate the driver/API config from the config file parsing. I found each project has slightly different config requirements, along with test code, that combining the too was pretty messy. This allows the projects to manage their own config file formatting for the options uniquely.

mongoclient/db.go
```diff
@@ -0,0 +1,106 @@
...
+
+type Config struct {
+	AppName            string
+	TLS                *nconf.TLSConfig
+	Servers            []string
+	ReplSetName        string
+	ConnTimeout        time.Duration
+	Auth               *Auth
+	SecondaryPreferred bool
+}
```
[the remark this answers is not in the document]
mheffner: Sure, will take a look

mongoclient/db.go
mheffner: I like that idea. Let me open an issue for that as that's somewhat a departure from the previous API, but I think a good direction to take.

mongoclient/db.go
rybit: I agree that the mashup is kinda ugly (do you add `mapstructure` or `split_words` etc) ~ but this means that each person would have to re-implement this struct & add their flags. Personally, I like the idea of "add this config to your struct, then add the env var" setup. So you don't have a lot of mapping code (`libconfig.SecondaryPreferred = mycode.SecondaryPreferred` kinda things) & we get more consistent naming.

Issue #146, opened 2020-08-11
Explore an Option driven configuration for modules


mheffner:
See the comment [here](https://github.com/netlify/netlify-commons/pull/144#discussion_r468853760), but it would be useful to have a consistent configuration approach to the various modules in this project. The options approach would also be a bit more forward-compatible as we may add options in the future.

PR #147, opened 2020-08-11
add options for the mongo client

rybit:
fixes https://github.com/netlify/netlify-commons/issues/146

I was thinking of adding struct tags around validation/mapstructure/envconfig to the `Config` object so that it is usable by default. Then if someone wants to roll their own config, they can use the `Option` interface.

### DIFF
--- a/mongoclient/db.go
+++ b/mongoclient/db.go
@@ -38,62 +38,29 @@ type Config struct {
 	SecondaryPreferred bool
 }
 
-// Connect connects to MongoDB
-func Connect(config *Config, log *logrus.Entry) (*mongo.Client, error) {
-	opts := options.Client().
-		SetConnectTimeout(config.ConnTimeout).
-		SetReplicaSet(config.ReplSetName).
-		SetHosts(config.Servers)
-
-	if config.AppName != "" {
-		opts.SetAppName(config.AppName)
-	}
-
-	if config.SecondaryPreferred {
-		opts.SetReadPreference(readpref.SecondaryPreferred())
-	}
-
-	if config.TLS != nil && config.TLS.Enabled {
-		tlsLog := log.WithFields(logrus.Fields{
-			"cert_file": config.TLS.CertFile,
-			"key_file":  config.TLS.KeyFile,
-			"ca_files":  strings.Join(config.TLS.CAFiles, ","),
-		})
-
-		tlsLog.Debug("Using TLS config")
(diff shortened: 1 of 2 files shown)


mongoclient/db.go
```diff
@@ -38,62 +38,29 @@ type Config struct {
...
-			"key_file":  config.TLS.KeyFile,
-			"ca_files":  strings.Join(config.TLS.CAFiles, ","),
-		})
-
-		tlsLog.Debug("Using TLS config")
-		tlsConfig, err := config.TLS.TLSConfig()
-		if err != nil {
+func ConnectWithOptions(log logrus.FieldLogger, replSet string, servers []string, opts ...Option) (*mongo.Client, error) {
+	finalOpts := options.Client().
+		SetConnectTimeout(time.Minute). // use a default
```
mheffner, line 43: We want a config option for conn timeout right?

mongoclient/db.go
```diff
@@ -38,62 +38,29 @@ type Config struct {
...
-	client, err := mongo.Connect(context.Background(), opts)
+	client, err := mongo.Connect(context.Background(), finalOpts)
 	if err != nil {
 		return nil, err
 	}
 
 	// Connect does not block for server discovery, so we should ping to ensure
 	// we are actually connected
-	ctx, cancel := context.WithTimeout(context.Background(), config.ConnTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), *finalOpts.ConnectTimeout+time.Second)
```
mheffner, line 63: Why the additional second?

rybit: So I was thinking about this. We set the `SetConnectTimeout` and then we use a timeout on the context with the exact same. I think that this could lead to some inconsistency in the errors. For instance, if the context times out vs the underlying lib timing out.

I guess this raises the question to me, do we need a timeout enforced here? Or should we rely on the timeout from the library?

mongoclient/db.go
rybit: This sets a sane default, I added an `Option` that will make this easy to override, but they can override it at any point (same with `replSet` and `servers` technically)

mongoclient/db.go
mheffner: This might be total request time vs. just a connect timeout. Thinking about it some more, we should probably separate the two timeouts.

rybit: If we run into this being a request or issue i think we can surface it then

PR #148, opened 2020-08-12
feat(kafka): add new package

smoya:
Package `kafka` provides a Consumer and a Producer for basic Kafka operations.

It relies on https://github.com/confluentinc/confluent-kafka-go which is a Go wrapper on top of https://github.com/edenhill/librdkafka.
This provides a reliable implementation, fully supported by the community, but also from Confluent, the creators of Kafka.

Once merge, generated **godoc** documentation including some examples will be available in [pkg.go.dev](https://pkg.go.dev/mod/github.com/netlify/netlify-commons?tab=packages).

### DIFF
--- a/kafka/config.go
+++ b/kafka/config.go
@@ -14,15 +14,15 @@ import (
 
 // Config holds all the configuration for this package.
 type Config struct {
-	Brokers      []string       `json:"brokers"`
-	Topic        string         `json:"topic"`
-	ProducerConf ProducerConfig `json:"producer" envconfig:"producer"`
-	ConsumerConf ConsumerConfig `json:"consumer" envconfig:"consumer"`
-	AuthType     string         `json:"auth" split_words:"true"`
-	User         string         `json:"user"`
-	Password     string         `json:"password"`
-	CAPEMFile    string         `json:"ca_pem_file"`
-	LogLevel     string         `json:"log_level" split_words:"true"`
+	Brokers   []string       `json:"brokers"`
+	Topic     string         `json:"topic"`
+	Producer  ProducerConfig `json:"producer" envconfig:"producer"`
+	Consumer  ConsumerConfig `json:"consumer" envconfig:"consumer"`
+	AuthType  string         `json:"auth" split_words:"true"`
+	User      string         `json:"user"`
+	Password  string         `json:"password"`
+	CAPEMFile string         `json:"ca_pem_file"`
+	LogLevel  string         `json:"log_level" split_words:"true"`
 }
 
 // baseKafkaConfig provides the base config that applies to both consumers and producers.
--- a/kafka/config.go
+++ b/kafka/config.go
@@ -1,0 +1,182 @@
+package kafka
+
+import (
+	"context"
+	"fmt"
+	"log/syslog"
+	"strings"
+	"time"
+
+	kafkalib "github.com/confluentinc/confluent-kafka-go/kafka"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// Config holds all the configuration for this package.
+type Config struct {
+	Brokers      []string       `json:"brokers"`
+	Topic        string         `json:"topic"`
+	ProducerConf ProducerConfig `json:"producer" envconfig:"producer"`
+	ConsumerConf ConsumerConfig `json:"consumer" envconfig:"consumer"`
+	AuthType     string         `json:"auth" split_words:"true"`
+	User         string         `json:"user"`
+	Password     string         `json:"password"`
+	CAPEMFile    string         `json:"ca_pem_file"`
+	LogLevel     string         `json:"log_level" split_words:"true"`
+}
+
+// baseKafkaConfig provides the base config that applies to both consumers and producers.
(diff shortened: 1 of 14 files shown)


kafka/config.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// Config holds all the configuration for this package.
+type Config struct {
+	Brokers      []string       `json:"brokers"`
+	Topic        string         `json:"topic"`
+	ProducerConf ProducerConfig `json:"producer" envconfig:"producer"`
+	ConsumerConf ConsumerConfig `json:"consumer" envconfig:"consumer"`
```
mraerino, line 20: these could be renamed since we no longer have methods with the names, right?

smoya:
@mraerino yeah, I think I'm gonna do it 👍

smoya:
@mraerino done!

kafka/example_test.go
```diff
@@ -0,0 +1,83 @@
...
+	defer c.Close()
+
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
+	// Consider implementing a retry mechanism.
+	for {
+		// 1. Fetch the message.
+		msg, err := c.FetchMessage(ctx)
+		if err != nil {
```
rybit, line 45: Is there a way to capture timeouts?

kafka/config.go
```diff
@@ -0,0 +1,192 @@
...
+	AuthType  string         `json:"auth" split_words:"true"`
+	User      string         `json:"user"`
+	Password  string         `json:"password"`
+	CAPEMFile string         `json:"ca_pem_file"`
+	LogLevel  string         `json:"log_level" split_words:"true"`
+}
+
+// baseKafkaConfig provides the base config that applies to both consumers and producers.
+func (c Config) baseKafkaConfig() *kafkalib.ConfigMap {
+	logrusToSylogLevelMapping := map[logrus.Level]syslog.Priority{
```
rybit, line 40: this could be pulled out into a var outside the method scope?

kafka/example_test.go
smoya: There is no timeout per se. The timeout inside the `FetchMessage` is a forced timeout, set in order to check the `ctx.Done` periodically.

kafka/config.go
smoya: What would be the benefit?

kafka/config.go
```diff
@@ -0,0 +1,192 @@
...
+
+// DefaultLogLevel is the log level Kafka producers/consumers will use if non set.
+const DefaultLogLevel = logrus.ErrorLevel
+
+// Config holds all the configuration for this package.
+type Config struct {
+	Brokers   []string       `json:"brokers"`
+	Topic     string         `json:"topic"`
+	Producer  ProducerConfig `json:"producer" envconfig:"producer"`
+	Consumer  ConsumerConfig `json:"consumer" envconfig:"consumer"`
```
rybit, line 30: I don't think you need the `envconfig` here, iirc it would mean that you'd have to specify it with _only_ that name and no nesting. 

e.g. for a config like:
```
type Config struct {
  Kafka kafka.Config
}
```
would need to be specified like

```
CONSUMER_GROUP_ID="blah"
```

versus 
```
KAFKA_CONSUMER_GROUP_ID="blah"
```

Please let me know if I'm wrong!

kafka/consume.go
```diff
@@ -0,0 +1,217 @@
...
+	kafkaConf := conf.baseKafkaConfig()
+	_ = kafkaConf.SetKey("enable.auto.offset.store", false) // manually StoreOffset after processing a message. Otherwise races may happen.)
+
+	// In case we try to assign an offset out of range (greater than log-end-offset), consumer will use start consuming from offset zero.
+	_ = kafkaConf.SetKey("auto.offset.reset", "earliest")
+
+	conf.Consumer.Apply(kafkaConf)
+	for _, opt := range opts {
+		opt(kafkaConf)
+	}
```
rybit, line 65: should we flip these? This makes it soe that you can't override things like `enable.auto.offset.store`.

kafka/config.go
smoya: Yeah, good point! needed it in the past because the field was called `ProducerConf` and `ConsumerConf`. Gonna change it right now.

kafka/consume.go
smoya: Not sure if I understand correctly. You can actually override the defaults by just setting an option. As it runs after applying the default option, you can override it.

kafka/config.go
smoya: Done!

kafka/config.go
rybit: you wouldn't re-instantiate it at every invocation, minor b/c this is only invoked once-ish.

rybit: Feel free to ignore

kafka/consume.go
rybit: Oh ~ wait. I read this wrong. I was thinking you were applying to something else. Nevermind. :)

PR #150, opened 2020-08-13
figure out why test is failing

rybit:


rybit:
ok looks like the test is flakey ~ I'll have to dig more into what it _should_ be doing.

vbrown608:
It comes from https://github.com/netlify/netlify-commons/pull/116, which we reviewed on the internal Bitballoon issue - see https://github.com/netlify/bitballoon/issues/5376#issuecomment-610548824

PR #158, opened 2020-08-20
Add murmurhash2 fucntion to kafka lib

phenixrizen:
## Summary
This PR adds the murmurhash2 function to the Kafka library.  This will be needed in socketeer to be able to hash a key to know what partition to read from.  The producer, filebeat will also have to hash.hash as the partition type.

## Test plan
Tests are provided.


phenixrizen:
I'm hoping this can be used with a PR that is open on the confluent lib https://github.com/confluentinc/confluent-kafka-go/pull/456

smoya:
I know its already merged but, can we can we add the right credits of the `murmur2` function to https://github.com/segmentio/kafka-go/blob/master/balancer.go#L253-L297 ?
Also we could almost copy paste (with few modifications) the tests they have in https://github.com/segmentio/kafka-go/blob/master/balancer_test.go#L141

@phenixrizen

PR #159, opened 2020-08-20
add a LogErrToSpan

rybit:
This method is used in a few services and it is pretty useful. I think that it makes sense to import here. Also, anything to make opentracing more usable I'm all for

### DIFF
--- a/tracing/opentracing.go
+++ b/tracing/opentracing.go
@@ -1,0 +1,24 @@
+package tracing
+
+import (
+	"fmt"
+
+	"github.com/opentracing/opentracing-go"
+	otlog "github.com/opentracing/opentracing-go/log"
+	"github.com/pkg/errors"
+)
+
+type stackTracer interface {
+	StackTrace() errors.StackTrace
+}
+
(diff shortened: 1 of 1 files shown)


tracing/opentracing.go
```diff
@@ -0,0 +1,24 @@
...
+	StackTrace() errors.StackTrace
+}
+
+func LogErrorToSpan(span opentracing.Span, err error) {
+	if err == nil || span == nil {
+		return
+	}
+
+	span.LogFields(otlog.String("event", "error"), otlog.Error(err))
+	if sterr, ok := err.(stackTracer); ok {
```
emilyzhang, line 21: out of curiosity what is happening here? O: it looks like you're checking to see whether you can create a new instance of the stacktracer interface from the error? (i've just never seen this `err.(stackTracer)` notation before

rybit: The general format is that you can do:
```
actualVersion, ifItMatches := someinterface.(whatYouWant)
```

This is using go's type inference to figure out if the error (which is an interface) passed in has that method.

PR #160, opened 2020-08-21
Support non .env files easily

rybit:
Based on a conversation in...a ticket that I can't find. It was noted that our infra already supports making _non .env_ files. This is really nice because it allows for other structured types. Specifically, netlify-server and proxy use json.

This makes it so that we look at the extension in setup. It also gets rid of a lot of the magic that we were doing with viper. I did a quick check with viper's .env support and it will just dump the values in as is - no map/slice support as best I can tell.

I also kept them as separate entry points `LoadFromX` to minimize the amount of change. I was debating making a single `LoadConfig` (like what existed) but decided against it. I'm open to changing that for sure


mheffner:
I think viper uses mapstructure where godotenv does not, I believe that means some of the struct attributes behave differently. We may need to be explicit on how we specify the key names and defaults in attributes?

rybit:
Hmmm @mheffner do you mean just adding some docs to the README? I tried adding examples in the tests, but you're right over communication is good.

mheffner:
> Hmmm @mheffner do you mean just adding some docs to the README? I tried adding examples in the tests, but you're right over communication is good.

More a question of whether this would cause a (silent) regression in something that assumes they work the next time someone did an updated of the nf-commons version?

rybit:
I’ll double check in the morning, but I think not. This should notice a .env and apply the same behavior. Other types should be cool with it and able to build from here

rybit:
So looking through it @mheffner - I think it won't. The cases I can see are:
- you use a `config.env` file (which I think most do). This should go by the same code path as before (through the godotenv/envconfig)
- you use an env file _without a .env_. This would make it try to parse as json/yaml. If you had no suffix it would bomb out and fail to start right away. Solution = add an extension to your file. I can't think of any services that are in this bucket
- you use json/yaml already. This should go through the same viper code to load (`LoadFromFile`). I was thinking of using json/yaml marshalers directly, but viper was already there.

PR #161, opened 2020-08-24
Adding produce and consume from partition capabilities to kafka lib

phenixrizen:
### Summary
- Added the ability to produce and consume from a partition.
- Added methods:
  - Consumer
    - GetMeta
    - GetPartitions
    - SetPartitionByKey
    - SeekToTime
  - Producer
    - GetMeta
    - GetPartitions
  - Config
    -WithPartitionerAlgorithm
- Added the docker-compose that launches a kafka cluster with 3 partitions

### Test plan
```
make integration-tests
```


kafka/consume.go
```diff
@@ -90,14 +95,45 @@ func (r *ConfluentConsumer) Seek(offset int64, timeout time.Duration) error {
 		return errors.New("Timeout should be set when calling Seek")
 	}
 
-	err := r.c.Seek(kafkalib.TopicPartition{Topic: &r.conf.Topic, Offset: kafkalib.Offset(offset)}, int(timeoutMs))
+	var err error
+	if r.partition < 0 {
+		err = r.c.Seek(kafkalib.TopicPartition{Topic: &r.conf.Topic, Offset: kafkalib.Offset(offset)}, int(timeoutMs))
+	} else {
+		err = r.c.Seek(kafkalib.TopicPartition{Topic: &r.conf.Topic, Partition: r.partition, Offset: kafkalib.Offset(offset)}, int(timeoutMs))
+	}
```
rybit, line 103: Like this without the white space errors I have

```suggestion
       topicPartition := kafkalib.TopicPartition{
               Topic: &r.conf.Topic, 
               Offset: kafkalib.Offset(offset),
               }
	 if r.partition >= 0 {
             topicPartition.Partition = r.partition
	  } 
	  err := r.c.Seek(topicPartition, int(timeoutMs))
	}
```

kafka/consume.go
```diff
@@ -90,14 +95,45 @@ func (r *ConfluentConsumer) Seek(offset int64, timeout time.Duration) error {
...
+
+	var offsets []kafkalib.TopicPartition
+	var err error
+	millisSinceEpoch := t.UnixNano() / 1000000
+	if r.partition < 0 {
+		offsets, err = r.c.OffsetsForTimes([]kafka.TopicPartition{{Topic: &r.conf.Topic, Offset: kafkalib.Offset(millisSinceEpoch)}}, int(timeoutMs))
+	} else {
+		offsets, err = r.c.OffsetsForTimes([]kafka.TopicPartition{{Topic: &r.conf.Topic, Partition: r.partition, Offset: kafkalib.Offset(millisSinceEpoch)}}, int(timeoutMs))
+	}
```
rybit, line 126: I'd do something similar to my above comment

kafka/consume.go
```diff
@@ -192,6 +240,41 @@ func (r *ConfluentConsumer) Close() error {
 	return r.c.Close()
 }
 
+// GetMeta return the confluence consumer metatdata
+func (r *ConfluentConsumer) GetMeta(allTopics bool, timeout time.Duration) (*kafkalib.Metadata, error) {
```
rybit, line 244: What do you think about setting a default timeout in the Consumer for operations, then having some kind of override? Instead of having it in each signature? 

I think that it might be a way to get rid of the boilerplate checks

phenixrizen: I actually flip flopped on this. I had it as a default setting in my last commit and then changed it to follow the other signatures in the library. I don't have any strong feelings either way￼

rybit: I think a default is my light preference.

kafka/consume.go
```diff
@@ -142,14 +178,26 @@ func (r *ConfluentConsumer) SetInitialOffset(offset int64) error {
...
+		})
+	} else {
+		parts := []kafkalib.TopicPartition{
+			kafkalib.TopicPartition{
+				Topic:     &r.conf.Topic,
+				Partition: r.partition,
+			},
 		}
-	})
+		err = r.c.Assign(parts)
```
smoya, line 198: You wanna assign just once.

kafka/consume.go
smoya: Also, can we change the name of the method to `GetMetadata` to be more friendly?

kafka/consume.go
```diff
@@ -74,9 +78,10 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
 	}
 
 	confluenceConsumer := &ConfluentConsumer{
-		c:    consumer,
-		conf: conf,
-		log:  log,
+		c:         consumer,
+		conf:      conf,
+		log:       log,
+		partition: -1,
```
smoya, line 84: would not be more idiomatic if the type is `*uint32`? so we can compare with `nil`?

Dockercompose.test.yml
```diff
@@ -0,0 +1,129 @@
...
+      - zookeeper-3
+    environment:
+      KAFKA_BROKER_ID: 3
+      KAFKA_ZOOKEEPER_CONNECT: localhost:22181,localhost:32181,localhost:42181
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://localhost:39092
+      KAFKA_CONFLUENT_SUPPORT_METRICS_ENABLE: "false"
+    extra_hosts:
+      - "moby:127.0.0.1"
+    logging:
+      driver: none  # change this to json-file if you want to debug kafka
```
mraerino, line 98: do you really need a 3-node cluster? i have other services where i used single-instance setups just fine

kafka/consume.go
phenixrizen: I was thinking about this last night.  I think we should add the sync once and initialize a new sync.Once when we set a new partition with SetPartitionByKey method.  That would allow a single connection to read from multiple partitions

Makefile
```diff
@@ -11,3 +11,10 @@ deps:
 
 test:
 	go test ./...
+
+integration-test:
+	docker-compose -f Dockercompose.test.yml up --build --abort-on-container-exit --always-recreate-deps
+	docker-compose -f Dockercompose.test.yml down --volumes
```
mraerino, line 17: you could replace these with a `docker-compose run golang_tests` so it automatically stops after running them

kafka/config.go
```diff
@@ -17,6 +17,14 @@ const (
...
 	AuthTypeSCRAM256 = "scram-sha256"
 	AuthTypeSCRAM512 = "scram-sha512"
+
+	PartitionerRandom           = "random"            // random distribution
+	PartitionerConsistent       = "consistent"        //  CRC32 hash of key (Empty and NULL keys are mapped to single partition)
+	PartitionerConsistentRandom = "consistent_random" // CRC32 hash of key (Empty and NULL keys are randomly partitioned)
+	PartitionerMurMur2          = "murmur2"           // Java Producer compatible Murmur2 hash of key (NULL keys are mapped to single partition)
+	PartitionerMurMur2Random    = "murmur2_random"    // Java Producer compatible Murmur2 hash of key (NULL keys are randomly partitioned. Default partitioner in the Java Producer.)
+	PartitionerFNV1A            = "fnv1a"             // FNV-1a hash of key (NULL keys are mapped to single partition)
+	PartitionerFNV1ARandom      = "fnv1a_random"      // FNV-1a hash of key (NULL keys are randomly partitioned).
```
mraerino, line 27: would be nice if you made a newtype over string and used it here, so this feels more like an enum and you can type arguments taking these better

kafka/config.go
```diff
@@ -190,3 +198,10 @@ func WithConsumerGroupID(groupID string) ConfigOpt {
 		_ = c.SetKey("group.id", groupID)
 	}
 }
+
+// WithPartitionerAlgorithm sets the partitioner algorithm
+func WithPartitionerAlgorithm(algorithm string) ConfigOpt {
```
mraerino, line 203: as said above, a specific type for `algorithm` would be awesome

kafka/consume.go
```diff
@@ -90,14 +95,45 @@ func (r *ConfluentConsumer) Seek(offset int64, timeout time.Duration) error {
...
+		return errors.New("Timeout should be set when calling Seek")
+	}
+
+	var offsets []kafkalib.TopicPartition
+	var err error
+	millisSinceEpoch := t.UnixNano() / 1000000
+	if r.partition < 0 {
+		offsets, err = r.c.OffsetsForTimes([]kafka.TopicPartition{{Topic: &r.conf.Topic, Offset: kafkalib.Offset(millisSinceEpoch)}}, int(timeoutMs))
+	} else {
+		offsets, err = r.c.OffsetsForTimes([]kafka.TopicPartition{{Topic: &r.conf.Topic, Partition: r.partition, Offset: kafkalib.Offset(millisSinceEpoch)}}, int(timeoutMs))
```
mraerino, line 125: what happens if the time is out of range of what kafka holds?
i remember there was an option to specify if you want to earliest or the latest offset in that case. something like that?

kafka/consume.go
mraerino: i'd support having a `sync.Once` for this :+1:

kafka/consume.go
phenixrizen: `-1` is a idiomatic way look at strings.Index or io.EOF

kafka/consume.go
```diff
@@ -192,6 +240,41 @@ func (r *ConfluentConsumer) Close() error {
...
+	return getPartitionIds(r.conf.Topic, meta)
+}
+
+// SetPartitionByKey sets the partition to consume messages from by the passed key
+func (r *ConfluentConsumer) SetPartitionByKey(key string, timeout time.Duration) error {
+	parts, err := r.GetPartions(timeout)
+	if err != nil {
+		return err
+	}
+	r.partition = GetPartition(key, parts)
```
mraerino, line 273: i understand this only works before you started consuming from kafka. it would be nice to have an atomic bool we flip once we assigned the partition and return an error here, if someone tries to set the partition afterwards

kafka/hash.go
```diff
@@ -1,12 +1,13 @@
...
-func GetPartition(msg *kafkalib.Message, partitions ...int) (partition int) {
+func GetPartition(key string, partitions []int32) int32 {
+	if len(partitions) == 0 {
+		return -1
+	}
 	// NOTE: the murmur2 balancers in java and librdkafka treat a nil key as
 	//       non-existent while treating an empty slice as a defined value.
-	idx := (murmur2(msg.Key) & 0x7fffffff) % uint32(len(partitions))
-	return partitions[idx]
+	idx := (murmur2([]byte(key)) & 0x7fffffff) % uint32(len(partitions))
```
mraerino, line 9: wouldn't you need to use the configured algorithm here?

Dockercompose.test.yml
phenixrizen: A single node would be fine if you truly didn't want to test the library properly.  I want to do some chaos testing, i.e. kill a container and make sure the library rebalances and still functions afterward.  We have not tested anything like this to my knowledge.

kafka/hash.go
phenixrizen: We will need to add additional hashing algorithms to support reading from a partition.  So you made me think we should track this and throw an error if the hashing balancer is not set to murmur2

kafka/consume.go
phenixrizen: The seek will return and error

kafka/consume.go
phenixrizen: What do you think would be a good default timeout, 30sec?  I will make it an option in the config?

rybit: Yeah I'd make it an option in the config, but default of 30s is plenty imho

kafka/consume.go
rybit: Is there actually a way that we could remove this method and take an option in the Constructor?

phenixrizen:
@smoya I had to move the assign and subscribe to the constructor because seeks would error if they were run before a read.  We really want to be able to seek before a read.

Lots of details in the last commit message for https://github.com/netlify/netlify-commons/pull/161/commits/9a3e7b56f9658703de5e4e3322b0f3307d2cf2e1

kafka/consume.go
```diff
@@ -73,29 +75,83 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
+	}
+
+	if cc.partition < 0 {
+		// if we are not reading from a specific partition, we subscribe
+		cc.configOnce.Do(func() {
+			cc.log.WithField("kafka_topic", cc.conf.Topic).Debug("Subscribing to Kafka topic")
+			cc.rebalanceHandlerMutex.Lock()
+			defer cc.rebalanceHandlerMutex.Unlock()
+			if err := cc.c.Subscribe(cc.conf.Topic, cc.rebalanceHandler); err != nil {
+				err = errors.Wrap(err, "error subscribing to topic")
```
rybit, line 101: This shadows the `err` I think

kafka/consume.go
```diff
@@ -73,29 +75,83 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
+	} else {
+		// if we are reading from a specific partition, we assign
+		cc.configOnce.Do(func() {
+			parts := []kafkalib.TopicPartition{
+				kafkalib.TopicPartition{
+					Topic:     &cc.conf.Topic,
+					Partition: cc.partition,
+				},
+			}
+			err = cc.c.Assign(parts)
```
rybit, line 113: what happens on a rebalance?

kafka/consume.go
```diff
@@ -73,29 +75,83 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
+	cc := &ConfluentConsumer{
+		c:         consumer,
+		conf:      conf,
+		log:       log,
+		partition: part,
+	}
+
+	if cc.partition < 0 {
+		// if we are not reading from a specific partition, we subscribe
+		cc.configOnce.Do(func() {
```
rybit, line 96: Why do we need a `Once` here? It seems like we aren't using it anywhere else than in this

kafka/consume.go
```diff
@@ -73,29 +75,83 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
-	confluenceConsumer := &ConfluentConsumer{
-		c:    consumer,
-		conf: conf,
-		log:  log,
+	if conf.Timeout == 0 {
+		conf.Timeout = DefaultTimout
 	}
 
-	return confluenceConsumer, nil
+	part := int32(-1)
```
smoya, line 82: The config is already a pointer. Don't understand why we need to change the type in here instead of use the same in both (pointer to).

kafka/consume.go
```diff
@@ -73,29 +75,83 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
+		log:       log,
+		partition: part,
+	}
+
+	if cc.partition < 0 {
+		// if we are not reading from a specific partition, we subscribe
+		cc.configOnce.Do(func() {
+			cc.log.WithField("kafka_topic", cc.conf.Topic).Debug("Subscribing to Kafka topic")
+			cc.rebalanceHandlerMutex.Lock()
+			defer cc.rebalanceHandlerMutex.Unlock()
```
smoya, line 99: This is not gonna work because nobody set the `cc.rebalanceHandler` yet via `SetInitialOffset()`.

smoya: THen it means also we are missing the testcase where we manually set the offset to start consuming from

kafka/consume.go
smoya: You mean the scope of `err` is just the If?

smoya:
@phenixrizen You don't want to use `Seek()` for setting the initial offset but `SetInitialOffset`. Read: https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/consumer.go#L230-L233

Either I would do all the `SetInitialOffset` via config (like you did with Partition but for Offset) or revert back to previous implementation.

Also the Seek is not fulfilling the interface but tests are still green? :\

Interface: `Seek(offset int64, timeout time.Duration) error`
Implementation: `func (r *ConfluentConsumer) Seek(offset int64) error {`

kafka/consume.go
phenixrizen: we will have a goroutine that listens to events and handles the rebalance logic

phenixrizen:
> Also the Seek is not fulfilling the interface but tests are still green? :\
> 
> Interface: `Seek(offset int64, timeout time.Duration) error`
> Implementation: `func (r *ConfluentConsumer) Seek(offset int64) error {`

This is because this code does not implement the interface.  We return a *ConfluentConsumer and not a OffsetAwareConsumer

rybit:
Personally, I'd get rid of the difference between Consumer and OffsetAwareConsumer - I don't find the distinction all that helpful right now. And then either return the interface or just add the nil declaration to the top of the file.

kafka/consume.go
rybit: yup!

phenixrizen:
So the Config now takes a Partition or a PartitionKey & PartitionerAlgorithm.  This allows us to rebalance if we know the partition key.  In the case of the just a parttion id being set, there is no way to know how to rebalance so the consumer will return errors when trying to fetch a message.

kafka/consume.go
```diff
@@ -79,39 +85,66 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
-		conf:      conf,
-		log:       log,
-		partition: part,
+		c:    consumer,
+		conf: conf,
+		log:  log,
 	}
 
-	if cc.partition < 0 {
+	if cc.conf.Consumer.Partition == nil {
```
rybit, line 94: does this mean that I'd have configure on a partition if I wanted to partition by a key? 
```
conf := ConsumerConfig {
  Partition: 1, // anything will do
  PartitionKey: "what I actually care about",
  PartitionAlgorithm: "actually have to set this", 
}
```

So I think if you did a 
```
if cc.conf.Consumer.Partition == nil && cc.conf.Consumer.PartitionKey == "" {
```

kafka/consume.go
```diff
@@ -79,39 +85,66 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
-					Partition: cc.partition,
-				},
+		tps := []kafkalib.TopicPartition{
+			kafkalib.TopicPartition{
+				Topic:     &cc.conf.Topic,
+				Partition: *cc.conf.Consumer.Partition,
+			},
+		}
+		// Set the partition if a key is set to determine the partition
+		if cc.conf.Consumer.PartitionKey != "" && cc.conf.Consumer.PartitionerAlgorithm != "" {
```
rybit, line 112: Could we default the `PartitionerAlgorithm`?

kafka/consume.go
```diff
@@ -79,39 +85,66 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
+					}
+					if revoked {
+						cc.log.WithField("kafka_event", e.String()).Debug("Unassigning Kafka partitions after rebalance")
+						if err := cc.c.Unassign(); err != nil {
+							log.WithError(err).Error("failed unassigning current Kafka partitions after rebalance")
+						}
+						// if we know the partition key we can reassign
+						if cc.conf.Consumer.PartitionKey != "" && cc.conf.Consumer.PartitionerAlgorithm != "" {
+							cc.SetPartitionByKey(cc.conf.Consumer.PartitionKey, cc.conf.Consumer.PartitionerAlgorithm)
+						}
```
rybit, line 143: What is going to happen if I am assigned to partition 1 and then that partition goes away?

kafka/consume.go
```diff
@@ -79,39 +85,66 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
+		if cc.conf.Consumer.PartitionKey != "" && cc.conf.Consumer.PartitionerAlgorithm != "" {
+			cc.SetPartitionByKey(cc.conf.Consumer.PartitionKey, cc.conf.Consumer.PartitionerAlgorithm)
+		}
+		err = cc.c.Assign(tps)
+
+		// handle rebalance events for consumer assigned to a specific partition
+		// - NOTE if the partition is set and the partition key is not set we have no way
+		//   of knowing where to assign the consumer to in the case of a rebalance
+		cc.eventChan = cc.c.Events()
+		go func(cc *ConfluentConsumer) {
```
rybit, line 121: Minor: we should pull this out into a method I think.

rybit:
some minor things, but largely I like it

kafka/consume.go
phenixrizen: So we only need the partition key and the partition algorithm

kafka/consume.go
phenixrizen: I was thinking we can do this in the constructor

kafka/consume.go
phenixrizen: The underlying c library will return a Kafka error.  There is no way to really know a partition to switch to.

kafka/consume.go
phenixrizen: sure

kafka/consume.go
rybit: oh wait ~ I thought this was all in the constructor 🤦

kafka/consume.go
rybit: will it cause the consumer to shutdown?

kafka/consume.go
rybit: So if we default the alg then key || id is what we would need. But if you set only key here you'd `Subscribe` and not do the `Assign`

kafka/consume.go
rybit: We are! but I don't see it

kafka/consume.go
```diff
@@ -73,33 +81,91 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
+	cc := &ConfluentConsumer{
 		c:    consumer,
 		conf: conf,
 		log:  log,
 	}
 
-	return confluenceConsumer, nil
+	if cc.conf.Consumer.Partition == nil {
+		// if we are not reading from a specific partition, we subscribe
+		cc.setupRebalanceHandler(cc.conf.Consumer.InitialOffset)
```
smoya, line 96: This should only be called if `cc.conf.Consumer.InitialOffset` is set.

kafka/consume.go
```diff
@@ -73,33 +81,91 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
+		cc.log.WithField("kafka_topic", cc.conf.Topic).Debug("Subscribing to Kafka topic")
+		cc.rebalanceHandlerMutex.Lock()
```
smoya, line 98: Why do you need Mutex here?

kafka/consume.go
```diff
@@ -73,33 +81,91 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
+	}
+	if len(offsets) == 1 {
+		return r.Seek(int64(offsets[0].Offset))
+	}
+
+	return fmt.Errorf("error finding offset to seek to")
+}
+
+// setupReabalnceHandler does the setup of the rebalance handler
+func (r *ConfluentConsumer) setupRebalanceHandler(offset int64) error {
```
smoya, line 168: The returned error is never different than `nil` so it can be removed.

kafka/consume.go
```diff
@@ -73,33 +81,91 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
 		return nil, err
 	}
 
-	confluenceConsumer := &ConfluentConsumer{
+	if conf.Timeout == 0 {
+		conf.Timeout = DefaultTimout
```
smoya, line 85:
```suggestion
		conf.Timeout = DefaultTimeout
```

kafka/config.go
```diff
@@ -12,11 +12,23 @@ import (
...
+
+	PartitionerRandom           = PartitionerAlgorithm("random")            // random distribution
+	PartitionerConsistent       = PartitionerAlgorithm("consistent")        //  CRC32 hash of key (Empty and NULL keys are mapped to single partition)
+	PartitionerConsistentRandom = PartitionerAlgorithm("consistent_random") // CRC32 hash of key (Empty and NULL keys are randomly partitioned)
+	PartitionerMurMur2          = PartitionerAlgorithm("murmur2")           // Java Producer compatible Murmur2 hash of key (NULL keys are mapped to single partition)
+	PartitionerMurMur2Random    = PartitionerAlgorithm("murmur2_random")    // Java Producer compatible Murmur2 hash of key (NULL keys are randomly partitioned. Default partitioner in the Java Producer.)
+	PartitionerFNV1A            = PartitionerAlgorithm("fnv1a")             // FNV-1a hash of key (NULL keys are mapped to single partition)
+	PartitionerFNV1ARandom      = PartitionerAlgorithm("fnv1a_random")      // FNV-1a hash of key (NULL keys are randomly partitioned).
+
+	DefaultTimout = time.Duration(30 * time.Second) // Default timout to be used if not set in the config
```
smoya, line 31:
```suggestion
	DefaultTimeout = time.Duration(30 * time.Second) // Default timeout to be used if not set in the config
```

kafka/config.go
```diff
@@ -33,6 +45,7 @@ type Config struct {
 	Password  string         `json:"password"`
 	CAPEMFile string         `json:"ca_pem_file"`
 	LogLevel  string         `json:"log_level" split_words:"true"`
+	Timeout   time.Duration  `json:"timeout"`
```
smoya, line 48: The field name `Timeout` is not clear enough. What timeout is referring to exactly? When configuring a Kafka Consumer or Producer you can set tons of them for several purposes. Can you rename the field for making the purpose obvious?

kafka/config.go
smoya: Also: https://github.com/netlify/netlify-commons/pull/161/files#r480039753

kafka/consume.go
```diff
@@ -73,33 +81,91 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
 
-	return confluenceConsumer, nil
+	if cc.conf.Consumer.Partition == nil {
+		// if we are not reading from a specific partition, we subscribe
+		cc.setupRebalanceHandler(cc.conf.Consumer.InitialOffset)
+		cc.log.WithField("kafka_topic", cc.conf.Topic).Debug("Subscribing to Kafka topic")
+		cc.rebalanceHandlerMutex.Lock()
+		defer cc.rebalanceHandlerMutex.Unlock()
+		if serr := cc.c.Subscribe(cc.conf.Topic, cc.rebalanceHandler); serr != nil {
+			err = errors.Wrap(serr, "error subscribing to topic")
```
smoya, line 101: Minor but you can do an early return here instead of delegating the error handling

kafka/consume.go
```diff
@@ -181,17 +262,80 @@ func (r *ConfluentConsumer) FetchMessage(ctx context.Context) (*kafkalib.Message
...
 	return errors.Wrap(err, "failed committing Kafka message")
 }
 
-// Close should be called when no more readings will be performed.
+// Close closes the consumer.
 func (r *ConfluentConsumer) Close() error {
 	return r.c.Close()
 }
 
+// GetMetadata return the confluence consumer metatdata
```
smoya, line 276:
```suggestion
// GetMetadata return the confluence consumer metadata
```

kafka/consume.go
```diff
@@ -118,33 +121,7 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (Consum
...
-						}
-						// if we know the partition key we can reassign
-						if cc.conf.Consumer.PartitionKey != "" && cc.conf.Consumer.PartitionerAlgorithm != "" {
-							cc.SetPartitionByKey(cc.conf.Consumer.PartitionKey, cc.conf.Consumer.PartitionerAlgorithm)
-						}
-					}
-				}
-			}
-		}(cc)
+		go cc.handlePartitionRebalacne()
```
smoya, line 124:
```suggestion
		go cc.handlePartitionRebalance()
```

kafka/consume.go
```diff
@@ -139,21 +205,36 @@ func (r *ConfluentConsumer) SetInitialOffset(offset int64) error {
...
-// FetchMessage implements Consumer interface.
-func (r *ConfluentConsumer) FetchMessage(ctx context.Context) (*kafkalib.Message, error) {
-	var err error
-	r.subscribeOnce.Do(func() {
-		r.log.WithField("kafka_topic", r.conf.Topic).Debug("Subscribing to Kafka topic")
-		r.rebalanceHandlerMutex.Lock()
-		defer r.rebalanceHandlerMutex.Unlock()
-		if err := r.c.Subscribe(r.conf.Topic, r.rebalanceHandler); err != nil {
-			err = errors.Wrap(err, "error subscribing to topic")
+func (r *ConfluentConsumer) handlePartitionRebalacne() {
```
smoya, line 208: This logic + `setupRebalanceHandler` can be managed in just one function.

kafka/consume.go
```diff
@@ -16,27 +16,32 @@ var ErrSeekTimedOut = errors.New("Kafka Seek timed out. Please try again.")
...
-	Consumer
+	// GetMetadata gets the metadata for a consumer.
+	GetMetadata(allTopics bool) (*kafkalib.Metadata, error)
 
-	// SetInitialOffset resets the current offset to the given one.
-	// Used for setting the initial offset a consumer should start consuming from.
-	// Should be called before start consuming messages.
-	SetInitialOffset(offset int64) error
+	// GetPartitions returns the partitions on the consumer.
+	GetPartions() ([]int32, error)
```
smoya, line 32:
```suggestion
	GetPartitions() ([]int32, error)
```

kafka/consume.go
```diff
@@ -73,33 +81,91 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
+			},
+		}
+		if cc.conf.Consumer.PartitionerAlgorithm == "" {
+			cc.conf.Consumer.PartitionerAlgorithm = PartitionerMurMur2
+		}
+		// Set the partition if a key is set to determine the partition
+		if cc.conf.Consumer.PartitionKey != "" && cc.conf.Consumer.PartitionerAlgorithm != "" {
+			cc.SetPartitionByKey(cc.conf.Consumer.PartitionKey, cc.conf.Consumer.PartitionerAlgorithm)
+		}
+		err = cc.c.Assign(tps)
```
smoya, line 118: I would return immediately in case of an error.

kafka/consume.go
phenixrizen: Good catch it was leftover

kafka/consume.go
```diff
@@ -73,33 +81,91 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*Confl
...
+
+		// handle rebalance events for consumer assigned to a specific partition
+		// - NOTE if the partition is set and the partition key is not set we have no way
+		//   of knowing where to assign the consumer to in the case of a rebalance
+		cc.eventChan = cc.c.Events()
```
smoya, line 123: Please read:

https://github.com/confluentinc/confluent-kafka-go/blob/d4b1fadc0776d02e9b74a1815f1e02dbb124fccd/kafka/kafka.go#L21-L57

kafka/consume.go
phenixrizen: The initial offset will be 0 and it will start to read there.  Am I missing something?

kafka/consume.go
phenixrizen: Good catch

kafka/consume.go
smoya: This is not the default behaviour of a Kafka consumer.  The default behaviour would be reading from the latest stored offset (if exists), otherwise will follow the strategy set in https://github.com/netlify/netlify-commons/blob/1b39178bb1f85791e2a13063b16285456455b673/kafka/consume.go (earliest).

kafka/consume.go
```diff
@@ -16,27 +16,32 @@ var ErrSeekTimedOut = errors.New("Kafka Seek timed out. Please try again.")
...
+	Seek(offset int64) error
+
+	// SeekToTime seeks to the specified time.
+	SeekToTime(t time.Time) error
+
+	// SetPartittionByKey sets the current consumer to read from a partion by a hashed key.
+	SetPartitionByKey(key string, algorithm PartitionerAlgorithm) error
+
+	// SetPartitionByID sets the current consumer to read from the specified partition.
+	SetPartitionByID(id int32) error
```
smoya, line 44: Can we skip this and move it out of the interface? I don't think it should be called from outside once the consumer is consuming.

Also a minor on the naming: I would prefer having more alignment in terms of naming with Kafka domain lang, so this could be named `AssignPartitionByID`, `AssignPartitionByKey` the other.

smoya: This would simplify the code and those who depend on it

smoya: As @phenixrizen explained to me via 📞 , this makes sense to stay there (perhaps just accepting my rename suggestion?) since this can be called once the consumer already started, making the consumer to switch to another partition without losing the TCP connection.

smoya:
I opened a new issue in order to track all the scenarios we want to support in this library regarding Consumers. The third scenario is the one **this** PR attempts to fulfill. https://github.com/netlify/netlify-commons/issues/164

phenixrizen:
I've added tests and the library still functions properly as it used to.  It looks like the events channel never gets set unless we don't poll. Even though there docks state otherwise. We should be able to use the subscribe and set the rebalance callback and set the partition to consume from the callback.  This will make the code much less complicated.

phenixrizen:
You can run `KAFKA_CHAOS=true make clean && make integration-test` and see the event channel is nil and no event is sent in the test because of this'

phenixrizen:
@smoya I will implement the change to have a single rebalance handler when I'm online tomorrow.  You should be able to start using this for the implementation of the log streaming tomorrow.  There will be no API changes. We can sync when I'm online.

kafka/consume.go
```diff
@@ -67,6 +68,13 @@ func NewConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (Consum
 	// In case we try to assign an offset out of range (greater than log-end-offset), consumer will use start consuming from offset zero.
 	_ = kafkaConf.SetKey("auto.offset.reset", "earliest")
 
+	// enable events channel for parttition consumer
+	if conf.Consumer.PartitionKey != "" {
+		defaultPartition := int32(0)
+		conf.Consumer.Partition = &defaultPartition
```
smoya, line 74: Why do we need to set a partition here?

kafka/consume.go
```diff
@@ -179,6 +186,7 @@ func (r *ConfluentConsumer) setupRebalanceHandler(offset int64) error {
 	var once sync.Once
 	r.rebalanceHandler = func(c *kafkalib.Consumer, ev kafkalib.Event) error {
 		log := r.log.WithField("kafka_event", ev.String())
+		spew.Dump(ev)
```
smoya, line 189: Friendly reminder to remove it

kafka/integration_test.go
```diff
@@ -10,79 +10,240 @@ import (
...
+func chaosTest(testBrokers string, assert *assert.Assertions) {
+	chaos := os.Getenv("KAFKA_CHAOS")
+	ctx := context.Background()
+	if chaos == "" {
+		a, err := kafkalib.NewAdminClient(&kafka.ConfigMap{"bootstrap.servers": testBrokers})
+		assert.NoError(err)
+		assert.NotNil(a)
+
+		maxDur, err := time.ParseDuration("60s")
+		assert.NoError(err)
```
smoya, line 236:
```suggestion
		maxDur := time.Second * 60
```

kafka/consume.go
phenixrizen: I've updated the logic to now consume at the offset provided by the broker if not initial offset is set in the config.

rybit (approved):

rybit (approved):

PR #166, opened 2020-09-01
user traffic log line parser

pandemicsyn:
Kick's off a parser package with a no frill's user traffic parser to start with. It's based on some previous work from @rybit in an old metering package. While this generally doesn't try to proactively clean up the log line in anyway - there was one case where we decided to clean up the result. Namely in that we proactively strip trailing comma's from `sid` , since this was a some what common occurrence.

### DIFF
--- a/parser/user_traffic.go
+++ b/parser/user_traffic.go
@@ -1,0 +1,141 @@
+package parser
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+)
+
+//UserTraffic is a single decoded user traffic log line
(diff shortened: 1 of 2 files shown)


pandemicsyn:
@rybit want me to just drop the benchmark from this pr, and send a long a separate pr with the benchmark code (and document a sample run in the pr) ? Or did you have something else in mind?

rybit:
Eh I'd just put it in a second PR

pandemicsyn:
👍  cool, dropped the benchmark from this change set.

PR #169, opened 2020-09-10
make the recoverer tie back to a request

rybit:
When debugging a panic this should introduce the requets that caused it and then a key we can use to sort from because humio messes that up.


mraerino (approved):
nice

emilyzhang (approved):

phenixrizen (approved):
Nice

PR #170, opened 2020-09-11
handle errors better in the tracing and spans

rybit:
In fighting the cert-store I noticed that when we fall into `HandleError` which is in the case of a panic or unexpected error we don't really cleanup nicely.

This adds that non-nil errors:
- unexpected ones go to bugsnag. This is because returning a 404 isn't an error in and of itself
- we always call `Finish` on the tracing. This is in the event of a panic
- we record unexpected errors in datadog
- we do complete the span if it isn't set


emilyzhang (approved):
neat

PR #173, opened 2020-09-21
Add traffic-mesh signature decoder.

mrdg:


trafficmesh/signature.go
```diff
@@ -0,0 +1,65 @@
...
+type SignaturePayload struct {
+	jwt.StandardClaims
+	SiteID    string `json:"sid,omitempty"`
+	DeployID  string `json:"did,omitempty"`
+	AccountID string `json:"aid,omitempty"`
+	URL       string `json:"url,omitempty"`
+	Remapped  bool   `json:"remapped,omitempty"`
+}
+
+func NewSignatureDecoder(secret string) *SignatureDecoder {
```
mraerino, line 27: this could use a comment that passing an empty string for secret is valid and will make `DecodeSignature` a no-op

mraerino (commented):
would be great if you documented the exported types

mraerino (approved):
cool!

Issue #164, opened 2020-08-31
kafka: Support all needed scenarios for Kafka consumers


smoya:
We want to support the following scenarios for Kafka consumers:

### Behavior:

- Consumer partitions are assigned automatically.
- Rebalance events are handled automatically by `librdkafka`.
- Consumer reads from latest stored offset. If none, earliest (could be changed). Configured [here](https://github.com/netlify/netlify-commons/blob/4b131f164bdfd1f17473f08edefcc602331f8fc5/kafka/consume.go#L60).

### Steps:

1. Create Consumer including `group.id`.
2. Call to `Subscribe` **without** a `Rebalance Callback` .
3. Start consuming.

### Behavior:

- Consumer partitions are assigned automatically.
- Rebalance events are handled locally in order to set the Initial offset after the first rebalance.
- Consumer reads from an specific offset.

### Steps:

1. Create Consumer including `group.id`.
2. Call to `Subscribe` and set a `Rebalance Callback` . In the first rebalance, this callback will assign the desired initial offset (calling `Assign`). Also de `Unassign` event should be handled gracefully.
3. Start consuming.

### Behavior:

- Consumer partitions are assigned **manually**.
- Several consumers can consume from the same partition.
- No rebalance events (not part of consumer group).

### Steps:

1. Create Consumer including a fake `group.id`. This is needed due to a limitation in `librdkafka`. See the following code in [confluent-kafka-go](https://github.com/confluentinc/confluent-kafka-go/blob/d4b1fadc0776d02e9b74a1815f1e02dbb124fccd/kafka/consumer.go#L386-L392).
2. **Never** call to `Subscribe`
3. Call to `Assign` with the desired partition (use whatever hashing algorithm you want for deciding which partition to read from).
4. Start consuming.

mraerino:
would be cool to split these scenarios into their own implementations so you can clearly select them by picking the correct reader impl.

rybit:
I think that it is interesting to have the use cases written out (thanks!) but I don't see the need to refactor this at this point. Right now we get a consumer that has a pretty standard signature, and we have plenty of configuration options that are intuitive to me. 

I think we should get #161 merged in and if we end up finding this to be very cumbersome we can revisit it. To me it would be refactoring just to refactor.

smoya:
The reason why I wrote those use cases is to set a guideline on what we need **now**. Have a clear path on what feature to work on and try to simplify the code we do as much as possible. 
[...]

PR #178, opened 2020-09-23
feat(kafka): add detached consumer

smoya:
Closes https://github.com/netlify/netlify-commons/issues/164

This PR adds a new constructor on the Kafka package for creating a Consumer detached from Consumer Groups.
Note the following (added in comments):

> A detached consumer will work out of consumer groups for partition assignment and rebalance, however it needs
>       permission on the group coordinator for managing commits, so it needs a consumer group in the broker.
>       In order to simplify, the default consumer group id is copied from the configured topic name, so make sure you have a
>       policy that gives permission to such consumer group.

Behaviour workflow follows recommendations from the creator of `librdkafka` - https://github.com/confluentinc/confluent-kafka-go/issues/534

### DIFF
--- a/kafka/consume.go
+++ b/kafka/consume.go
@@ -56,6 +56,75 @@ type ConfluentConsumer struct {
 	eventChan chan kafkalib.Event
 }
 
+// NewDetachedConsumer creates a Consumer detached from Consumer Groups for partition assignment and rebalance (see NOTE).
+// - NOTE Either a partition or partition key is required to be set.
+//       A detached consumer will work out of consumer groups for partition assignment and rebalance, however it needs
+//       permission on the group coordinator for managing commits, so it needs a consumer group in the broker.
+//       In order to simplify, the default consumer group id is copied from the configured topic name, so make sure you have a
+//       policy that gives permission to such consumer group.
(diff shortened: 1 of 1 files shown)


kafka/consume.go
```diff
@@ -176,8 +245,8 @@ func (cc *ConfluentConsumer) setupRebalanceHandler(offset *int64) {
 				// if we have an initial offset we need to set it
 				if cc.conf.Consumer.InitialOffset != nil {
 					once.Do(func() {
-						log.WithField("kafka_offset", *offset).Debug("Skipping Kafka assignment given by coordinator after rebalance in favor of resetting the offset")
-						partitions = kafkalib.TopicPartitions{{Topic: &cc.conf.Topic, Offset: kafkalib.Offset(*offset)}}
+						log.WithField("kafka_offset", *cc.conf.Consumer.InitialOffset).Debug("Skipping Kafka assignment given by coordinator after rebalance in favor of resetting the offset")
```
smoya, line 248: Took the opportunity to clean up the rebalance handler by using the InitialOffset from the config directly.

phenixrizen (approved):
LGTM

PR #179, opened 2020-09-23
Expose LD users to enable custom attributes.

mheffner:
If we want to test features against custom user attributes, we need be able to build a more complex user object when passing to the Enabled and Variation checks. This PR adds `*User` methods to expose the same checks, but instead of taking a simple string key takes a LD `User` instead. This allows the consumer to use the [`UserBuilder`](https://pkg.go.dev/github.com/launchdarkly/go-server-sdk#UserBuilder) to build a complex user object with attributes.

This does mean that we are now exposing the LD client internals directly to consumers, if they want to use these new APIs. At the moment I think this is fine as most consumers will stick to simple key-based methods only. This will allow us to experiment with more custom user attributes and build common patterns. We can eventually fold those best practices back into this library in a way that could maintain the abstraction. However, general rule-of-3, I think it's worth waiting until that point.

### DIFF
--- a/featureflag/featureflag.go
+++ b/featureflag/featureflag.go
@@ -10,7 +10,10 @@ import (
 
 type Client interface {
 	Enabled(key, userID string) bool
+	EnabledUser(key string, user ld.User) bool
+
 	Variation(key, defaultVal, userID string) string
+	VariationUser(key string, defaultVal string, user ld.User) string
 }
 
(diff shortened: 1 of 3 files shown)


mheffner:
I think the `go.sum` changes appeared because an update didn't commit them previously?

rybit (approved):

PR #180, opened 2020-09-24
fix(kafkaclient): do not close logs channel

smoya:
Logs channel was closed when the context got canceled. This ends in a race condition where the `confluent-kafka-go` library tries to still send log messages (client not closed yet) to a closed channel, producing a  panic: `panic: send on closed channel`

### DIFF
--- a/kafka/config.go
+++ b/kafka/config.go
@@ -1,7 +1,6 @@
 package kafka
 
 import (
-	"context"
 	"fmt"
 	"log/syslog"
 	"strings"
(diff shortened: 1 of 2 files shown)


smoya:
In fact, I'm thinking of removing the context support... and keep logging always (goroutine will be killed when the main app is terminated). In that way, we will capture all possible logs.

mrdg (approved):
makes sense to me!

smoya:
Removed context support. Now It's a breaking change, but it is under control 👍

kafka/config.go
```diff
@@ -175,27 +174,24 @@ func WithLogger(ctx context.Context, log logrus.FieldLogger) ConfigOpt {
 
 		// Read from channel and print logs using the provided logger.
 		go func() {
-			defer close(logsChan)
+			// Do not close logsChan because confluent-kafka-go will send logs until we close the client.
+			// Otherwise it will panic trying to send messages to a closed channel.
 			for {
-				select {
-				case <-ctx.Done():
+				m, ok := <-logsChan
```
mrdg, line 180: You could range over the channel as well instead.

mrdg (approved):

kafka/config.go
smoya: Sure, good catch! fixed

mrdg (approved):

PR #182, opened 2020-09-29
chore(ci): improve Docker Compose files for a more robust experience running integration tests

smoya:
Integration tests run on Docker via Docker Compose.
There was some issues on the configuration, on how we were creating the kafka topic for testing the `kafka` package, and some other minors (including using version 2 which is outdated).

This PR fixes all known issues and updates the Docker Compose file but also simplifies the configuration for Kafka (only need 1 zookeeper vs 3, removing host-network mode, etc).

It also fixes the integration test.

### DIFF
--- a/Dockercompose.test.yml
+++ b/Dockercompose.test.yml
@@ -1,116 +1,77 @@
 ---
-version: '2'
+version: '3'
+
 services:
-  zookeeper-1:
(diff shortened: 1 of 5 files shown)


mraerino (approved):
looks much cleaner!

PR #183, opened 2020-09-29
provide default config for logging

rybit:
I was trying to use this for a silly script, I didn't configure any logging parameters and then this started to panic. This should prevent that

### DIFF
--- a/nconf/args.go
+++ b/nconf/args.go
@@ -18,30 +18,24 @@ type RootArgs struct {
 	ConfigFile string
 }
 
+type RootConfig struct {
+	Log         LoggingConfig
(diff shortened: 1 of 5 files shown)


emilyzhang (approved):

rybit:
ok so I was working on this a bit more - but I was struggling with Viper, essentially it was doing the right mapping, but it wouldn't unmarshal properly. This gets rid of that dependency.

rybit:
This should be transparent:
- there are no services that use yaml to my knowledge
- the json parsing should work exactly the same
- no changes for env files

emilyzhang (approved):

PR #189, opened 2020-10-01
trace requests per route

rybit:
Essentially when I was looking at the tracing in cert-store (and others) all requests for the different routes are under 1 header ("GET" or "PUT") which make it really hard to figure out what's going on. This is an attempt to break that up into per pattern routes.

I figured that I'd have to ditch the middleware approach because it doesn't have the context to know what path it is. If we use the path from the request itself, we're going to get a huge number and they won't aggregate. Sooo this is my compromise.

It is a breaking change in that it changes the behaviour. I added the `TraceAllRequests` to the router's middleware if people want to upgrade but not change functionality. I will have to test that out though. It is also breaking in the API in that the `NewTracer` and `WrapWithSpan` now take a `resource`. I searched GH in the org and couldn't find any references to those methods outside of commons.

My plan is to:
- [x] integrate this branch with cert-store and see if I get the trace shape I want
- [ ] test to see if I can maintain behaviour with a few lines
- [x] discuss and see if this is something we want to adopt

If anyone sees a way that is less hackey to get this to work please let me know.


rybit:
ok I tested this out with cert-store and it did what I wanted. Take a look [here](https://app.datadoghq.com/apm/resource/cert-store/http.handler/4fd6a8baf73534c5?end=1601674975656&env=staging&hostGroup=%2A&index=apm-search&paused=true&query=env%3Astaging%20service%3Acert-store%20operation_name%3Ahttp.handler%20resource_name%3Aget.certificates.domain.resolve&start=1601674647205). 

image


The thing that is "fixed" is that the endpoints are now not just GET or PUT. They're now `method.endpoint`. This is so that you can have multiple endpoints with different HTTP methods but the same path and they show up as different endpoints.

mheffner (approved):
Nice improvement!

How about splitting the method from the resource name a bit more clearly? From bitballoon's APM, it looks like they maintain case, so maybe some variant of: `GET-resourcename`, `GET resourcename`, `GET::resourcename`

mheffner (approved):
:woman_dancing:

mheffner (dismissed):
Was the LD change required for this or just trying to make it less noisy?

rybit:
Not necessary - I am actually considering moving it to a separate PR but
wanted to see what the whole experience would be.

router/errors.go
```diff
@@ -86,40 +86,40 @@ func httpError(code int, fmtString string, args ...interface{}) *HTTPError {
...
+			notifyBugsnag = true
+			log.WithError(e).Infof("unexpected error: %s", e.InternalMessage)
 		}
 
 		if jsonErr := SendJSON(w, e.Code, e); jsonErr != nil {
 			log.WithError(jsonErr).Error("Failed to write the JSON error response")
 		}
 	default:
-		// this is 5ns slower than using unsafe but a unhandled internal server error should not happen that often
-		if reflect.ValueOf(err).IsNil() {
```
mheffner, line 120: Does moving this up to the top add latency we should be concerned about?

mheffner (approved):
looks good :woman_dancing:

router/errors.go
rybit: I don't think so. Also in the error cases I'm generally less worried about latency, everything is screwed already 😂

PR #192, opened 2020-10-21
feat(kafka): add ConsumerFactory in order to align future developments

smoya:
In an effort to help users of our Kafka lib with common use cases, this PR adds a Factory function type, which can be used for postponing the creation of a consumer on runtime. E.g. setting an initial offset that you don't know on boot time.

`NewConsumer` and `NewDetachedConsumer` implements this already.

### DIFF
--- a/kafka/consume.go
+++ b/kafka/consume.go
@@ -14,6 +14,11 @@ import (
 // ErrSeekTimedOut is the error returned when a consumer timed out during Seek.
 var ErrSeekTimedOut = errors.New("Kafka Seek timed out. Please try again.")
 
+// ConsumerFactory creates a Consumer.
+// You can use it for postponing the creation of a consumer on runtime. E.g. setting an initial offset that you don't know on boot time.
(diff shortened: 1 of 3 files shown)


kafka/consume.go
```diff
@@ -62,7 +67,7 @@ type ConfluentConsumer struct {
 //       permission on the group coordinator for managing commits, so it needs a consumer group in the broker.
 //       In order to simplify, the default consumer group id is copied from the configured topic name, so make sure you have a
 //       policy that gives permission to such consumer group.
-func NewDetachedConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (*ConfluentConsumer, error) {
+func NewDetachedConsumer(log logrus.FieldLogger, conf Config, opts ...ConfigOpt) (Consumer, error) {
```
smoya, line 70: This bc is under control 👍

mrdg (approved):

PR #196, opened 2020-12-04
Add regex based kv extraction option

pandemicsyn:
Changes the behaviour of the UT parser to collect any unknown kv pairs or unbalanced fields as a slice of unparsed entries (returned in the UserTraffic struct) rather than throwing an error. This is especially helpful since we somewhat frequently have duplicate value fields for keys.

For example - while auditing a sample of log lines for a Peloton site nearly 100% of the "key with no value" errors where from a country code field where the value was a duplicate like `US, US`. 

```
2020/12/04 08:29:30 ==== Errors Seen ====
2020/12/04 08:29:30 key with no value: 6341 
2020/12/04 08:29:30 missing_sid_and_csid: 1926
2020/12/04 08:29:30 lines seen: 19058
2020/12/04 08:29:30 bytes potentially billed: 611126576
2020/12/04 08:29:30 bytes potentially missed: 45379903
```

### DIFF
--- a/parser/user_traffic.go
+++ b/parser/user_traffic.go
@@ -2,9 +2,15 @@ package parser
 
 import (
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
+)
+
+var (
+	// regex to try and extract key=value entries
+	re = regexp.MustCompile(`(?m)(\w+)=([^\s]+)`)
 )
 
 //UserTraffic is a single decoded user traffic log line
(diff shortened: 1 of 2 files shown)


rybit:
I think that this is an ok solution, I'm wondering if it might be easier to actually not error on the case of a key with no value. We could just log a warning (or a counter?) and then move on?

rybit (dismissed):

pandemicsyn:
Yea - we have that catch all "other" map as the default case for when we run into unknown kv pair's: 

https://github.com/netlify/netlify-commons/blob/d1232d6210495872813a867018419ea67d936f70/parser/user_traffic.go#L154-L158

I tested a variation that just adds the offending "key with no value" error entries as key's with an empty value to that `others` map.  Then folks using the parser can check that map for entries and generate log lines/metrics accordingly. I'd prefer that to add logging/metrics to the parser directly. I'm definitely down for doing that instead of the regex. @rybit wdyt ?

rybit:
yeah I think that approach works nicely. Or maybe we change the 'other' to be a list of unparsable sections, the whole bit.

rybit (approved):

PR #197, opened 2020-12-09
Set status code as string on spans

kschlarman:
For whatever reason, in order to use the `http.status` tag of a trace span from a datadog dashboard, the status must be written as a string,  not an integer.

See [slack conversation](https://netlify.slack.com/archives/C9D3B8RGF/p1607535957411300).
See https://github.com/netlify/gotrue/pull/272 for equivalent working change in gotrue.

### DIFF
--- a/tracing/req_tracer.go
+++ b/tracing/req_tracer.go
@@ -2,6 +2,7 @@ package tracing
 
 import (
 	"net/http"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


vbrown608 (approved):
:thread:

rybit (approved):

mraerino:
fixing in https://github.com/netlify/netlify-commons/pull/201

PR #201, opened 2020-12-15
Fix tests around status code in tracing

mraerino:
Follow-up for #197


kschlarman (approved):
Or sorry! Thanks for fixing it.

Issue #199, opened 2020-12-15
Fix launchdarkly client logging


mraerino:
> LaunchDarkly seems to be logging with its own format in proxy logs, making the lines not parseable in Humio
>
> We should either stop logging this, log it to some other file or format them correctly
>
> (@jose-ledesma)

from https://github.com/netlify/proxy/issues/433

mraerino:
seems to have been fixed in https://github.com/netlify/netlify-commons/pull/189 🤔

PR #202, opened 2020-12-16
Fix ld logging config

mraerino:
Guess i do too much Rust to not expect implicit copying of values...

_Fixes https://github.com/netlify/netlify-commons/issues/199_


smoya (approved):
LGTM!

fasterthanlime:
> Guess i do too much Rust to not expect implicit copying of values...

Woops, same here, totally missed this last review.

PR #203, opened 2020-12-16
Fix usage of struct tags for config

mraerino:
#183 introduced a breaking change that meant the struct tag `mapstructure` could no longer be used to map config keys.

since this is used in various services we need to restore this behaviour.


nconf/configuration.go
```diff
@@ -24,15 +35,26 @@ func LoadFromFile(configFile string, input interface{}) error {
...
+	case ".yaml", ".yml":
+		err = yaml.Unmarshal(data, &config)
+	default:
+		err = &ErrUnknownConfigFormat{configExt}
+	}
+	if err != nil {
+		return fmt.Errorf("failed to read config: %w", err)
+	}
+
+	if err := mapstructure.Decode(&config, &input); err != nil {
```
mheffner, line 54: Should this be just `input`?

mraerino: mapstructure wants a pointer to a map

mraerino: ah, yup. changed.

mheffner (approved):
:+1:

PR #206, opened 2020-12-23
make the config direct from yaml and json

rybit:
Ok the old code was doing a bytes -> map[string]interface{} -> mapstructure. This was very confusing (and didn't seem to be working?) so this actually just simplifies it to be json/yaml directly.

What it means is that anyone using this will need to add the json/yaml tags to their struct (and can lose the mapstructure). I think that it makes it a bit clearer

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -52,7 +52,6 @@ require (
 	gopkg.in/ghodss/yaml.v1 v1.0.0 // indirect
 	gopkg.in/launchdarkly/go-sdk-common.v1 v1.0.0-20200401173443-991b2f427a01 // indirect
 	gopkg.in/launchdarkly/go-server-sdk.v4 v4.0.0-20200729232655-2a44fb361895
-	gopkg.in/yaml.v2 v2.2.8
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 )
 
--- a/go.mod
+++ b/go.mod
@@ -52,6 +52,7 @@ require (
 	gopkg.in/ghodss/yaml.v1 v1.0.0 // indirect
 	gopkg.in/launchdarkly/go-sdk-common.v1 v1.0.0-20200401173443-991b2f427a01 // indirect
 	gopkg.in/launchdarkly/go-server-sdk.v4 v4.0.0-20200729232655-2a44fb361895
+	gopkg.in/yaml.v2 v2.2.8
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 )
 
--- a/nconf/args.go
+++ b/nconf/args.go
@@ -18,15 +18,17 @@ type RootArgs struct {
 	ConfigFile string
 }
 
+type RootConfig struct {
+	Log         LoggingConfig
+	BugSnag     *BugSnagConfig
+	Metrics     metriks.Config
+	Tracing     tracing.Config
+	FeatureFlag featureflag.Config
+}
+
 func (args *RootArgs) Setup(config interface{}, serviceName, version string) (logrus.FieldLogger, error) {
 	// first load the logger and BugSnag config
-	rootConfig := &struct {
-		Log         *LoggingConfig
-		BugSnag     *BugSnagConfig
-		Metrics     metriks.Config
-		Tracing     tracing.Config
-		FeatureFlag featureflag.Config
-	}{}
+	rootConfig := &RootConfig{}
 
 	loader := func(cfg interface{}) error {
 		return LoadFromEnv(args.Prefix, args.ConfigFile, cfg)
--- a/nconf/args.go
+++ b/nconf/args.go
@@ -30,16 +30,7 @@ func (args *RootArgs) Setup(config interface{}, serviceName, version string) (lo
 	// first load the logger and BugSnag config
 	rootConfig := &RootConfig{}
 
-	loader := func(cfg interface{}) error {
-		return LoadFromEnv(args.Prefix, args.ConfigFile, cfg)
-	}
-	if !strings.HasSuffix(args.ConfigFile, ".env") {
-		loader = func(cfg interface{}) error {
-			return LoadFromFile(args.ConfigFile, cfg)
-		}
-	}
-
-	if err := loader(rootConfig); err != nil {
+	if err := args.load(rootConfig); err != nil {
 		return nil, errors.Wrap(err, "Failed to load the logging configuration")
 	}
 
@@ -73,7 +64,7 @@ func (args *RootArgs) Setup(config interface{}, serviceName, version string) (lo
 
 	if config != nil {
 		// second load the config for this project
-		if err := loader(config); err != nil {
+		if err := args.load(config); err != nil {
 			return log, errors.Wrap(err, "Failed to load the config object")
 		}
 		log.Debug("Loaded configuration")
--- a/nconf/args.go
+++ b/nconf/args.go
@@ -79,6 +79,18 @@ func (args *RootArgs) Setup(config interface{}, serviceName, version string) (lo
 	return log, nil
 }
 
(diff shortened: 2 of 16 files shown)


mheffner (commented):
Could we keep the existing `LoadFromFile` as is and make a new method ( eg `LoadFromConfig` (?)) that has this new behavior. I'm concerned that projects simply updating netlify-commons may inadvertently have a behavior change due to this.

Not all the individual module config files in this project that have the mapstructure tags seem to be updated with json/yaml tags. I think we need to update all of them in this to avoid unparsed options?

rybit:
So I updated to keep that in place, but use LoadConfigFromFile in the Setup call. I think that this might have some minor impacts, but should be transparent (🤞). 

What I'd like to do is eventually do an audit of services for what version of commons they're using and then do an update & deploy so we can remove the `mapstructure` tags everywhere.

mraerino:
i merged #204 so you likely need to rebase onto that being reverted again.

i fear this change has the risk of bringing a lot of unexpected breakage to different services. can you provide a plan for how to handle this change in the different services?
i don't like the prospect of services being blocked on upgrading commons because of this. this will impact developer velocity because of the monolithic nature of this module.

rybit:
ok so I was also worried about this, so I tosed together a simple script that pulled down and checked the repos. What I came up with was [this data](https://docs.google.com/spreadsheets/d/1C9fpd_-S3b1T1kN_XASL5M5W9xtpz8siNBbtPvfXqGU/edit#gid=1319418355). What it showed was that these services _don't_ use .env file configurations (well, not _infra's structure) and use commons:



repo | sha | branch | commons | envfile
-- | -- | -- | -- | --
blobsync | master | 8b125c8 | github.com/netlify/netlify-commons v0.36.1-0.20200806192201-858f35be3da9 | dne
buildbot | master | 036bf47 | github.com/netlify/netlify-commons v0.45.1 | dne
buildbot-cleanup-operator | master | b6c62a1 | github.com/netlify/netlify-commons v0.9.0 | dne
cachecontroller | master | e9ad0d1 | github.com/netlify/netlify-commons v0.19.1-0.20190910175506-300241bba69b | dne
consul-discovery | master | 48125c7 | github.com/netlify/netlify-commons v0.9.0 | dne
herald | master | 1365a3d | github.com/netlify/netlify-commons v0.36.1-0.20200802204407-212ffb66f54e | dne
netlify-server | master | 9da96b4 | github.com/netlify/netlify-commons v0.47.0 | dne
ns1controller | master | 8a7bb3c | github.com/netlify/netlify-commons v0.35.0 | dne
pinger | main | 4d13f86 | github.com/netlify/netlify-commons v0.48.4-0.20201223010053-148bbf69c2f3 | dne
proxy | metriks-logging-cleanup | f406188 | github.com/netlify/netlify-commons v0.48.4-0.20201228193817-125dd35dccdc | dne

Not a problem:
- blobsync: this uses the .env file, but it is in docker/config.env.tpl
- herald: this uses a .env file, but it is config.env.tpl
- buildbot: this is deployed largely through voodoo. It uses a configmap in kube and then a base64 encoded payload to do the rest of the config. 
- buildbot-cleanup-operator: very old, any update is going to have a lot of breaking changes. I think it is also stable and not under active development. 

Needs updates to confirm they're fine:
- cachecontroller: very old, but important, I think that this needs an update PR
- consul-discovery: this uses a json config file, it would need an update & redeploy (or deprecation)
- netlify-server: not too old, but needs an update PR. 
- ns1controller: this is a bit older. 
- pinger: this is a new service, it will get a simple update and then a simple redeploy
- proxy: this is this service

The perk is that these are pretty easy to test in an update. If they start & log & metric then they should be safe to deploy (famous last words).

mheffner:
> What I came up with was

looks we've got some work to do for converting everything to yaml :trollface:

rybit:
ok in that case - I'm thinking I'll make those PRs against this PR, show that it loads in test for them, then I can merge/tag commons & update the dep in those PRs

nconf/bugsnag.go
```diff
@@ -8,28 +8,28 @@ import (
 
 type BugSnagConfig struct {
 	Environment    string
```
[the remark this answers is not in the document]
rybit: both yaml & json support case insensitive unmarshalling. You only need the yaml/json tags when you're (1) outputting to a case sensitive format (2) using something like snake case. 

https://play.golang.org/p/Q4BN7N3bnZZ

nconf/args.go
```diff
@@ -97,6 +95,17 @@ func (args *RootArgs) MustSetup(config interface{}, serviceName, version string)
 	return logger
 }
 
+func (args *RootArgs) loadDefaultConfig() (*RootConfig, error) {
+	c := &RootConfig{
+		Log: DefaultLoggingConfig(),
+	}
+
+	if err := args.load(c); err != nil {
+		return nil, errors.Wrap(err, "Failed to load the logging configuration")
```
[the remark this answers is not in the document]
rybit: This is using the input from the config file that is passed in. So if you've got invalid yaml/json/env then it would tank. This surfaces that to the `Setup` and then `MustSetup` will panic on that error

nconf/bugsnag.go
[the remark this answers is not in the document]
mraerino: according to that playground the yaml examples didn't work though?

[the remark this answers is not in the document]
rybit: Wow. I read right past that. I definitely was like "yes that works". 

It turns out that the yaml lib actually automatically downcases _every key_. https://github.com/go-yaml/yaml/issues/123

Meaning for yaml, I'd have to actually make it `yaml:"Environment"`, which would break the lowercase version.

go.mod
```diff
@@ -52,6 +52,7 @@ require (
 	gopkg.in/ghodss/yaml.v1 v1.0.0 // indirect
 	gopkg.in/launchdarkly/go-sdk-common.v1 v1.0.0-20200401173443-991b2f427a01 // indirect
 	gopkg.in/launchdarkly/go-server-sdk.v4 v4.0.0-20200729232655-2a44fb361895
+	gopkg.in/yaml.v2 v2.2.8
```
mraerino, line 55: why add this? v3 doesn't work?

nconf/args_test.go
```diff
@@ -1,10 +1,14 @@
...
 
 import (
+	"encoding/json"
 	"io/ioutil"
+	"os"
 	"testing"
+	"time"
 
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
```
mraerino, line 11: maybe update to v3 here and remove the newline after this?

nconf/configuration.go
```diff
@@ -1,15 +1,61 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 
+	"github.com/netlify/netlify-commons/featureflag"
+	"github.com/netlify/netlify-commons/metriks"
+	"github.com/netlify/netlify-commons/tracing"
```
mraerino, line 13: if you want to split these off, they usually go below other packages. i wish gofmt enforced this

nconf/configuration_test.go
```diff
@@ -9,13 +9,16 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
```
mraerino, line 12: version upgrade

mraerino (commented):
no big things. these should be quick to fix. will be there for approval

nconf/configuration.go
rybit: yes I wish that it would do that automatically. I just lumped them together for the time being so fmt can order them nicely

mraerino (approved):

PR #212, opened 2021-02-10
Add functions and edge_handlers to netlify toml

vbrown608:
Adding fields permitted in [https://docs.netlify.com/configure-builds/file-based-configuration/#sample-file](file-based configuration).


rybit:
I was thinking about this a bit when I saw the PR. I don't think that there is much sharing the toml file right now, it might make more sense to have this live in the buildbot for the time being. It will also mean you're less susceptible to churn on the other stuff.

ehmicky (approved):

Issue #208, opened 2021-01-07
Support DataDog Distribution metric


mrdg:
https://godoc.org/github.com/DataDog/datadog-go/statsd#Client.Distribution
https://docs.datadoghq.com/metrics/distributions/

> Global distributions are designed to instrument logical objects, like services, independently from the underlying hosts. Unlike histograms which aggregate on the Agent-side, global distributions send all raw data collected during the flush interval and the aggregation occurs server-side

This seems to be the recommended way to record quantiles across different servers, as opposed to the [function](https://godoc.org/github.com/DataDog/datadog-go/statsd#Client.TimeInMilliseconds) we're using currently (which calculates percentiles on the agent and can't be aggregated afterwards). The Distribution api doesn't seem to be supported in armon/go-metrics, perhaps because it can't be supported in the other sinks. We could consider using the datadog package directly in netlify-commons/metriks.

rybit:
I think moving off the armon version might be what we need to do. 😭 We also want to have an ability to report a value on a timer (e.g. keep a gauge fresh, or report only the last value on a period). 

I did do a quick and dirty integration with distribution metrics: https://github.com/netlify/pinger/blob/e4350a2843a277a9d1de24e6a1c3fb41e9ccbda1/report/report.go#L38-L60

Also, there is a very old version of a metrics package in commons: https://github.com/netlify/netlify-commons/tree/633daeda658f0244b152f5e2df7130c67d1a0811/metrics

kitop:
Coming to add a +1 for this. It'd be really helpful to graph the subscription delay in analytics.

PR #216, opened 2021-03-04
build in distribution metrics for datadog sink

rybit:
fixes: https://github.com/netlify/netlify-commons/issues/208

I decided to do this for only the datadog sink because I think it is where it is most useful. It means other sinks will throw this data away.

Also, yes it is icky global where concurrent invocations would be..odd. We could fix this with some locking but I thought the risk was low enough to not worry about that. Part of me was thinking of storing the `statsd` globally but the other supporting values (e.g. default tags, service name) would have to go with it. If we decide to cut the metrics package we can reconsider this whole thing.

### DIFF
--- a/metriks/distribution_test.go
+++ b/metriks/distribution_test.go
@@ -31,5 +31,5 @@ func TestDistribution(t *testing.T) {
 	Distribution("some_metric", 12.0, L("b", "c"))
 
 	buf := <-msgs
-	assert.Equal(t, "some_metric.some_metric:12|d|#a:1,b:c", string(bytes.Trim(buf, "\x00")))
+	assert.Equal(t, "testing.some_metric:12|d|#a:1,b:c", string(bytes.Trim(buf, "\x00")))
 }
--- a/metriks/distribution_test.go
+++ b/metriks/distribution_test.go
@@ -1,0 +1,35 @@
+package metriks
+
+import (
+	"bytes"
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDistribution(t *testing.T) {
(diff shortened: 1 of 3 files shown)


metriks/distribution_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	require.NoError(t, initDistribution(pc.LocalAddr().String(), "testing", []string{"a:1"}))
+	SetDistributionErrorHandler(func(err error) {
+		assert.NoError(t, err)
+	})
+
+	Distribution("some_metric", 12.0, L("b", "c"))
+
+	buf := <-msgs
+	assert.Equal(t, "some_metric.some_metric:12|d|#a:1,b:c", string(bytes.Trim(buf, "\x00")))
```
mheffner, line 34: Does this add the service name prefix if constructed from the `metriks.Init` pkg?

rybit: arg ~ i messed up the variable scoping thing. Should be fixed

rybit: the actual prefixing happens deep in the go-metrics statsd sink

metriks/metriks.go
```diff
@@ -104,7 +104,9 @@ func createDatadogSink(url string, name string, tags map[string]string, extraTag
 	}
 
 	sink.SetTags(ddTags)
-
+	if err := initDistribution(url, name, ddTags); err != nil {
```
mheffner, line 107: It doesn't look like this `name` field is the serviceName?

rybit: wow - yeah this whole init section is a bit of a mess. But that is a separate thing. Let me fix that (that's what I get for making some assumptions!)

mheffner (approved):
:shipit:

PR #217, opened 2021-03-04
make a persistent gauge

rybit:
In a handful of services we've needed a gauge that would report on some scheduled timer. This does that so that we can consistently report sparse gauge values to datadog

### DIFF
--- a/metriks/gauge.go
+++ b/metriks/gauge.go
@@ -12,6 +12,10 @@ const (
 	defaultGaugeDuration = time.Second * 10
 )
 
+// PersistentGauge will report on an interval the value to the metrics collector.
+// Every call to the methods to modify the value immediately report, but if we
+// don't have a change inside the window (default 10s) after the last report
+// we will report the current value.
 type PersistentGauge struct {
 	name  string
 	value int32
@@ -22,17 +26,21 @@ type PersistentGauge struct {
 	dur    time.Duration
 }
 
+// Set will replace the value with a new one, it returns the old value
 func (g *PersistentGauge) Set(value int32) int32 {
 	v := atomic.SwapInt32(&g.value, value)
 	g.report(value)
 	return v
 }
 
+// Inc will +1 to the current value and return the new value
 func (g *PersistentGauge) Inc() int32 {
 	v := atomic.AddInt32(&g.value, 1)
 	g.report(v)
 	return v
 }
+
+// Dec will -1 to the current value and return the new value
 func (g *PersistentGauge) Dec() int32 {
 	v := atomic.AddInt32(&g.value, -1)
 	g.report(v)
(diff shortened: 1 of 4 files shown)


metriks/gauge.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (g *PersistentGauge) start(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-g.ticker.C:
+			g.report(g.value)
+		}
```
mheffner, line 62: What do you think about using the ScheduledExecutor util here?

metriks/gauge.go
```diff
@@ -0,0 +1,88 @@
...
+
+func NewGaugeWithDuration(name string, tags []metrics.Label, dur time.Duration) *PersistentGauge {
+	ctx, cancel := context.WithCancel(context.Background())
+	g := PersistentGauge{
+		name:   name,
+		tags:   tags,
+		ticker: time.NewTicker(dur),
+		cancel: cancel,
+		dur:    dur,
+	}
```
mheffner, line 85: What do you think about having this take a callback function that computes the value at the time of report? A user could register this with a function that reads the gauge value (eg. current length of queue) and we'd invoke it once every 10secs and report as a gauge. That could ensure you always have a reading every interval.

metriks/gauge.go
rybit: I was thinking about that ~ one thing that I was seeing is that I couldn't do a 'Reset' on it. What I was intending on doing with this was that it would only report a value, if we hadn't reported in x seconds.

metriks/gauge.go
rybit: Interesting, I kinda like it. But one thing that I'd worry about is that this value is actually being buffered outside of us by the agent. So if we're reporting every 10s, that doesn't mean that we're in sync with that windowing. The way that this is done we report every value to the agent (so high throughput ones would be caught there). We'd also lose the Inc/Dec features. 

I am thinking I'll make another gauge 'ScheduledGauge' that will do this. Seems really useful

metriks/gauge.go
mheffner: Oh I get this now. So, maybe an alternative would be don't report on each operation, just record the value, then the timer would report the current value every 10secs evenly?

metriks/gauge.go
```diff
@@ -0,0 +1,115 @@
...
+	return NewScheduledGaugeWithDuration(name, defaultGaugeDuration, cb, tags...)
+}
+
+// NewScheduledGaugeWithDuration will create an start a ScheduledGauge that reports the value every period
+func NewScheduledGaugeWithDuration(name string, dur time.Duration, cb func() int32, tags ...metrics.Label) ScheduledGauge {
+	g := ScheduledGauge{
+		ScheduledExecutor: util.NewScheduledExecutor(dur, func() {
+			v := cb()
+			Gauge(name, float32(v), tags...)
+			fmt.Println("sent the value?")
```
mheffner, line 110: todo? :-)

metriks/gauge.go
rybit: I made the `ScheduledGauge` that would do this. But in even testing it I found some oddness in the sense of "when is it actually reported". I think having 2 of them could be nice, one that will internally track the value and the second that will use a callback

metriks/gauge.go
```diff
@@ -0,0 +1,113 @@
...
+	"context"
+	"sync/atomic"
+	"time"
+
+	"github.com/armon/go-metrics"
+	"github.com/netlify/netlify-commons/util"
+)
+
+const (
+	defaultGaugeDuration = time.Second * 10
```
mheffner, line 13: So if I look at our custom metrics, it looks like most of them report on a 10sec frequency. So best practice sampling says we should measure at N/2 or 5 secs to avoid the case where we don't report in a given interval. I realize this is configurable, but maybe best to start with a default that would work in our case?

mheffner (approved):
Minor comment on default duration, but :100:

metriks/gauge.go
rybit: good call ~ updated!

mheffner (approved):
:shirt:

PR #218, opened 2021-03-09
build a test logger

rybit:
I was waiting for some humio queries to finish and decided that I'd finally write this. What the goal is to make it easier to inject a logger all over our tests.

There is a lot of
```
func tl(t *testing.T) logrus.FieldLogger {
  return logrus.WithField("test", t.Name())
}
```

And `init` blocks that check for the log level in a variety of forms. This is an effort to have to never copy/paste that code again. While I was there I also added some features I think are nice, but I'm open to removing them.

### DIFF
--- a/testutil/logger.go
+++ b/testutil/logger.go
@@ -11,6 +11,12 @@ import (
 // Opt is a function that will modify the logger used
 type Opt func(l *logrus.Logger)
 
+// TL will build and return a test logger
+func TL(t *testing.T, opts ...Opt) *logrus.Entry {
+	l, _ := TestLogger(t, opts...)
+	return l
+}
+
 // TestLogger will build a logger that is useful for debugging
 // it respects levels configured by the 'LOG_LEVEL' env var.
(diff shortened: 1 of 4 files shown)


testutil/logger.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/sirupsen/logrus/hooks/test"
+)
+
+// Opt is a function that will modify the logger used
+type Opt func(l *logrus.Logger)
+
+// TestLogger will build a logger that is useful for debugging
+// it respects levels configured by the 'LOG_LEVEL' env var.
+// It takes opt functions to modify the logger used
+func TestLogger(t *testing.T, opts ...Opt) (*logrus.Entry, *test.Hook) {
```
mheffner, line 17: I feel personally there have been few times I'd need the log hook when building a test logger. In terms of brevity like your example, maybe there could be an option that just returned a log entry without the hook?

mheffner (approved):
:shipit:

Issue #177, opened 2020-09-22
[server] Add WriteTimeout and ReadTimeout to server configuration. Set defaults if apply.


smoya:
Server package should allow configuring both `WriteTimeout` and `ReadTimeout`. We can evaluate setting a default option as well.

```
    // WriteTimeout is the maximum duration before timing out
    // writes of the response. It is reset whenever a new
    // request's header is read. Like ReadTimeout, it does not
    // let Handlers make decisions on a per-request basis.
```

```
    // ReadTimeout is the maximum duration for reading the entire
    // request, including the body.
    //
    // Because ReadTimeout does not let Handlers make per-request
    // decisions on each request body's acceptable deadline or
    // upload rate, most users will prefer to use
    // ReadHeaderTimeout. It is valid to use them both.
```

PR #219, opened 2021-03-09
take optional Opts for modifying the http server

rybit:
fixes https://github.com/netlify/netlify-commons/issues/177

This should also be extensible for other settings as they come up

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -52,7 +52,8 @@ type HealthChecker interface {
 	Healthy(w http.ResponseWriter, r *http.Request) error
 }
 
-func New(log logrus.FieldLogger, config Config, api APIDefinition) (*Server, error) {
+// New will build a server with the defaults in place
+func New(log logrus.FieldLogger, config Config, api APIDefinition, opts ...Opt) (*Server, error) {
 	var healthHandler router.APIHandler
 	if checker, ok := api.(HealthChecker); ok {
 		healthHandler = checker.Healthy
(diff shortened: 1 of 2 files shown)


mraerino (commented):
looks like we're now mixing two ways to configure the server:
- use the `Config` struct
- use the opts

I'd like to use one or the other not both.
Generally i'm in favor of the opts approach it's close to a builder pattern, but more flexible in Go.

Maybe consider making a different constructor that doesn't take `Config`, but mostly opts and people can decide which they want to use?

rybit:
Good idea - I'll do that. I was thinking that things in Config would be considered "required" and things in "Opts" would be additional modifications on top of that.

mraerino:
yeah, was thinking about that. since in general we should try to make less things required and provide more opinionated defaults we should try to make all required things explicit parameters imho

rybit:
I did something where it kinda unified them a bit. One pure Opts and one with config. I am not in love with it :/

server/server.go
```diff
@@ -52,44 +56,30 @@ type HealthChecker interface {
 	Healthy(w http.ResponseWriter, r *http.Request) error
 }
 
-func New(log logrus.FieldLogger, config Config, api APIDefinition) (*Server, error) {
-	var healthHandler router.APIHandler
-	if checker, ok := api.(HealthChecker); ok {
-		healthHandler = checker.Healthy
+func NewOpts(log logrus.FieldLogger, api APIDefinition, opts ...Opt) (*Server, error) {
```
keiko713, line 59: THIS IS YOUR TIME TO SHINE AND PUT THE COMMENT HERE, RYAN

mraerino (dismissed):
doesn't look too bad to me 👍

mraerino (approved):

PR #221, opened 2021-03-12
Fix FakeKafkaConsumer to advance through messages

kschlarman:
**Summary**
I seems the the `FakeKafkaConsumer` will only advance to read the next message if you commit the offset, which isn't how a real consumer is expected to operate.

I added an `uncommitedOffset` to the consumer which will advance after every `FetchMessage`.

### DIFF
--- a/kafka/kafkatest/utils.go
+++ b/kafka/kafkatest/utils.go
@@ -125,6 +125,7 @@ func (f *FakeKafkaConsumer) CommitMessage(msg *kafkalib.Message) error {
 func (f *FakeKafkaConsumer) SetInitialOffset(offset int64) error {
 	f.msgMu.Lock()
 	f.offset = offset
+	f.uncommitedOffset = offset
 	f.msgMu.Unlock()
 	return nil
 }
@@ -132,6 +133,7 @@ func (f *FakeKafkaConsumer) SetInitialOffset(offset int64) error {
 func (f *FakeKafkaConsumer) Seek(offset int64) error {
 	f.msgMu.Lock()
 	f.offset = offset
+	f.uncommitedOffset = offset
 	f.msgMu.Unlock()
 	return nil
 }
--- a/kafka/kafkatest/utils.go
+++ b/kafka/kafkatest/utils.go
@@ -33,13 +33,13 @@ func KafkaPipe(log logrus.FieldLogger) (*FakeKafkaConsumer, *FakeKafkaProducer) 
 }
 
 type FakeKafkaConsumer struct {
-	messages         []*kafkalib.Message
-	msgMu            sync.Mutex
-	offset           int64
-	uncommitedOffset int64
-	notify           chan struct{}
-	commits          chan *kafkalib.Message
-	log              logrus.FieldLogger
+	messages   []*kafkalib.Message
+	msgMu      sync.Mutex
+	offset     int64
+	readOffset int64
+	notify     chan struct{}
+	commits    chan *kafkalib.Message
+	log        logrus.FieldLogger
 }
 
 type FakeKafkaProducer struct {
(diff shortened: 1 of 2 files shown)


kafka/kafkatest/utils.go
```diff
@@ -86,11 +88,12 @@ func NewFakeKafkaConsumer(log logrus.FieldLogger, distri <-chan *kafkalib.Messag
...
 		f.msgMu.Lock()
-		if int64(len(f.messages)) > f.offset {
-			f.log.WithField("offset", f.offset).Trace("offering message")
-			msg := f.messages[f.offset]
+		if int64(len(f.messages)) > f.uncommitedOffset {
+			f.log.WithField("offset", f.uncommitedOffset).Trace("offering message")
+			msg := f.messages[f.uncommitedOffset]
 			f.msgMu.Unlock()
 
+			f.uncommitedOffset = f.uncommitedOffset + 1
```
mheffner, line 96: This seems unrelated to your PR, but do we need to bump `offset` here: https://github.com/netlify/netlify-commons/blob/a8a24e784a013eb660f38b141e33c6d7585f12c7/kafka/kafkatest/utils.go#L76 ?

kafka/kafkatest/utils.go
```diff
@@ -63,11 +64,12 @@ func (f *FakeKafkaProducer) Close() error {
...
 func NewFakeKafkaConsumer(log logrus.FieldLogger, distri <-chan *kafkalib.Message) *FakeKafkaConsumer {
 	r := &FakeKafkaConsumer{
-		messages: make([]*kafkalib.Message, 0),
-		offset:   0,
-		notify:   make(chan struct{}),
-		log:      log,
-		commits:  make(chan *kafkalib.Message, 1000),
+		messages:         make([]*kafkalib.Message, 0),
+		offset:           0,
+		uncommitedOffset: 0,
```
mheffner, line 69: "uncommitted" seems wrong if this is tracking the read unconnected to the committed offsets now?, `readOffset` ?

kafka/kafkatest/utils_test.go
```diff
@@ -15,17 +15,27 @@ func TestFakeKafkaProducer_WaitForKey(t *testing.T) {
...
-		Key:   key,
+	ctx := context.Background()
+
+	err := p.Produce(ctx, &kafkalib.Message{
+		Key:   []byte(`key1`),
+		Value: []byte(`val1`),
+	})
+	err = p.Produce(ctx, &kafkalib.Message{
+		Key:   []byte(`key2`),
+		Value: []byte(`val2`),
```
rybit, line 26: you're not checking the errors along the way, it seems like we should

kafka/kafkatest/utils.go
```diff
@@ -108,6 +111,7 @@ func (f *FakeKafkaConsumer) CommitMessage(msg *kafkalib.Message) error {
 	f.log.WithField("offset", msg.TopicPartition.Offset).Trace("commiting message...")
 	if int64(msg.TopicPartition.Offset) > f.offset {
 		f.offset = int64(msg.TopicPartition.Offset)
+		f.uncommitedOffset = f.offset
```
mheffner, line 114: Would this cause the reader to jump back to this msg then? It feels like this could loop on the same msg

kafka/kafkatest/utils.go
```diff
@@ -121,13 +125,15 @@ func (f *FakeKafkaConsumer) CommitMessage(msg *kafkalib.Message) error {
...
 	f.offset = offset
+	f.uncommitedOffset = offset
 	f.msgMu.Unlock()
 	return nil
 }
 
 func (f *FakeKafkaConsumer) Seek(offset int64) error {
 	f.msgMu.Lock()
 	f.offset = offset
+	f.uncommitedOffset = offset
```
mheffner, line 136: Should these two methods now "not" set the `offset`?

mheffner (commented):
Largely questions because I'm less familiar what the intended behavior of this consumer is.

kafka/kafkatest/utils.go
kschlarman: I think the intent was that the `offset` was the committed offset, so that would not be touched unless we called `CommitOffset` or `Seek`. The linked code is setting the offset for incoming messages that have not been fetched yet.

kafka/kafkatest/utils.go
kschlarman: Oh good point

kafka/kafkatest/utils.go
kschlarman: Yah, I'm not super good with Kafka terminology. If read makes more sense, I'm 👍 .

kschlarman:
I've got some tests up [in WIP PR](https://github.com/netlify/log-drains/pull/40/files#diff-fa535ef5e370b74ca5d2f6bb1edbdadfff34b27e9907a8db83bf5aa0f706d7ecR61) that shows this working.

mheffner (approved):
:+1:

PR #223, opened 2021-03-18
Add pull request template

kitop:
Initial very basic PR template. This is to start having better history, and start driving towards conventional commits so we can auto-document version changes.

It's a very raw version so please add feedback


rybit (approved):

mraerino:
when i'll add release-please i'm going to point out that we only need to follow conventional commits for PR titles

kitop:
Thanks! Feel free to edit the template when doing so!

PR #224, opened 2021-03-19
feat: allow setting initial delay & custom ticker in the scheduled executor

emilyzhang:
- adds custom Ticker interface and types to allow users to pass in their own ticker type or specify initial delay (including having the ticker start immediately) 
- should not change/break any existing behavior for scheduled executor (if people are expecting scheduled executor to wait for a duration before executing)

### DIFF
--- a/util/scheduled_executor.go
+++ b/util/scheduled_executor.go
@@ -90,7 +90,9 @@ func (s *scheduledExecutor) Stop() {
 		return
 	}
 
-	s.ticker.Stop()
+	if s.ticker != nil {
+		s.ticker.Stop()
+	}
 	s.done <- true
 	s.wg.Wait()
 
@@ -100,7 +102,20 @@ func (s *scheduledExecutor) Stop() {
 
 func (s *scheduledExecutor) poll() {
 	defer s.wg.Done()
-	time.Sleep(s.initialDelay)
+
+	// pause for initial delay
+	begin := make(chan struct{})
+	go func() {
+		time.Sleep(s.initialDelay)
+		begin <- struct{}{}
+	}()
+	select {
+	case <-s.done:
+		return
+	case <-begin:
+	}
+
+	// infinite loop for scheduled execution
 	for ; true; <-s.ticker.C() {
 		select {
 		case <-s.done:
--- a/util/scheduled_executor.go
+++ b/util/scheduled_executor.go
@@ -5,18 +5,11 @@ import (
 	"time"
 )
 
-type ScheduledExecutor interface {
-	Start()
-	Stop()
-}
-
 type CustomTicker interface {
 	Stop()
 	Start()
 	C() <-chan time.Time
 }
-
-type Option func(*scheduledExecutor)
 
 type defaultTicker struct {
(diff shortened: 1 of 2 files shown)


mheffner (commented):
What about making the initial delay time configurable? Idea stolen from: [`scheduleWithFixedDelay`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleWithFixedDelay(java.lang.Runnable,%20long,%20long,%20java.util.concurrent.TimeUnit))

emilyzhang:
@mheffner love this idea, thanks! i'll make some edits

mheffner:
Actually, it may be closer to [`scheduleAtFixedRate`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleAtFixedRate(java.lang.Runnable,%20long,%20long,%20java.util.concurrent.TimeUnit))

util/scheduled_executor.go
```diff
@@ -11,12 +11,13 @@ type ScheduledExecutor interface {
...
-	period    time.Duration
-	cb        func()
-	isRunning AtomicBool
-	ticker    *time.Ticker
-	done      chan bool
-	wg        sync.WaitGroup
+	period             time.Duration
+	cb                 func()
+	isRunning          AtomicBool
+	ticker             *time.Ticker
```
cbosss, line 17: possible thought... what if we could pass in a configurable ticker. 

ex: 

```go
type Ticker interface {
	Stop()
	Reset()
	C()
}
```

We could then provide some common used tickers in this package (PlainTicker, PlainTickerInstantStart...), but it would allow the calling service to create their own (maybe I want a JitteringTicker, or a ticker NonLinearTicker)...

emilyzhang: i love this idea!

mheffner:
Love where this is going!

util/scheduled_executor.go
```diff
@@ -5,26 +5,75 @@ import (
...
+		opt(s)
+	}
+	return s
+}
+
+func WithInitialDelay(initialDelay time.Duration) Option {
+	return func(s *scheduledExecutor) {
+		s.initialDelay = initialDelay
+	}
+}
```
cbosss, line 72: Now that we can specify a custom ticker... do we need to directly support this? Couldn't I just make a ticker that has an initial delay?

emilyzhang: that's a good point! i just figure if it's going to be used by multiple services, it might be easier to already have it specified here

cbosss: Initial delay complicates the implementation of this scheduled executor a bit. It would be nice if we could separate out that functionality into a custom ticker (which could be provided in this package). 

That way all ScheduledExecutor needs to know is that it will receive a ticker, and that it should run the executor when that thing ticks.

emilyzhang:
closing for now as i don't have the bandwidth to work on this at the moment. i might reopen again later (or anyone else should feel free to take this on if it is needed)

PR #226, opened 2021-03-25
feat: add instrument package for Segment

keiko713:
This PR adds the instrument package for Segment. You can enable this by having the following in the config file:

```
INSTRUMENT_ENABLED=true
INSTRUMENT_KEY=segment_api_key
```

Then, you can use this like the following:

```
instrument.Track("user_id", "my-event", properties)
```

If enabled is `false` (default `false`), you can still use this, but instead of the events sent out to Segment, it'll be logged to the logger.


instrument/instrument_test.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func TestLogging(t *testing.T) {
+	cfg := Config{
+		Key: "ABCD",
+	}
+
+	logBuf := new(bytes.Buffer)
+	log := logrus.New()
+	log.Out = logBuf
```
rybit, line 40: One thing you might consider using is in the testutil package there is a hook that is good for testing.

pandemicsyn:
I think it might be worth including some docs pointing netlify folks to https://github.com/netlify/segment-events to make sure internal users follow the processes and conventions mentioned there? 

I like the visibility of having this package in netlify-commons, because this is probably the first place folks would look for it, but I wonder if this shouldn't be a package in the internal segment events repo?

instrument/global.go
```diff
@@ -0,0 +1,35 @@
...
+	globalLock.Lock()
+	globalClient = client
+	globalLock.Unlock()
+}
+
+func GetGlobalClient() Client {
+	globalLock.Lock()
+	defer globalLock.Unlock()
+	return globalClient
+}
```
rybit, line 25: so because `Client` is exposed, I think you're going to have some issues with pushing people to one or another. I think that it might be better to actually make the `Client` interface private, expose its methods directly in the package (like featureflag). Then you'd have a function like this:

```
func Identify(...blah...) {
  getGlobalClient().Indetify(blah)
}
```

It might be good to keep it exposed so people can know, but have package level functions make it slighlyt better ergonomics. For instance, logrus, metriks, and featureflag do something akin to this.

keiko713: cool will update 👍 (fyi featureflag is not doing that as I copy&pasted this from featureflag :P)

keiko713:
> I think it might be worth including some docs pointing netlify folks to https://github.com/netlify/segment-events to make sure internal users follow the processes and conventions mentioned there?

In general, I like the idea of pointing to the doc. I think we can also improve our README to explain how to use each package too, and pointing to that repo is something we can do in there. How does that sound? I'm honestly little hesitate to add it to the codebase as a comment, so feels better if it was in README.

> I like the visibility of having this package in netlify-commons, because this is probably the first place folks would look for it, but I wonder if this shouldn't be a package in the internal segment events repo?

Sooooo I'd disagree with having this package in segment-events repo. netlify-commons repo is providing the basic common features of our services and the instrumentation can be count as a basic feature, was my reasoning having this here. For example, we could implements more instrumentation client potentially in the future and swap out as the project needs.
The netlify/segment-events is also not really fitting to have the code/library in it, I feel like. It is really more of the documentation.
I see this instrument package a very similar to the featureflag package. We are providing the basic feature in the netlify-commons library, then we have documented our own rule for how to use it elsewhere.

instrument/global.go
keiko713: actually nvm, I just found `func Enabled(key, userID string) bool` in featureflag package 🤦 apparently there is something I missed copy&pasta

keiko713:
uhmm re: doc, actually maybe I should be adding more in the README but more in the code (as you can see it in https://pkg.go.dev/github.com/netlify/netlify-commons@v0.51.2/metriks).
I'll actually add more doc in the code and I'll see how it turns out

keiko713:
doc output:
image

instrument/global.go
```diff
@@ -0,0 +1,61 @@
...
+var globalClient client = MockClient{}
+
+func setGlobalClient(client client) {
+	if client == nil {
+		return
+	}
+	globalLock.Lock()
+	globalClient = client
+	globalLock.Unlock()
+}
```
rybit, line 20: So minor things, but by hiding this method in the package, you make it so that people can't inject their own clients to the global space (e.g. they want to instrument the existing one). 

I think exporting this (and the `client` interface) might be good. Yes I know this is kinda conflicting my earlier comment, but having a global methods like you've got here are the real benefit I was trying to allude to

Same with `getGlobalClient` ~ as you hid the actually global client you can't save it off in tests. In the featureflag we do a lot of 
```
func TestSomething (t *testing.T) {
  og := featureflag.GetGlobalClient()
  t.Cleanup(func(){ featureflag.SetGlobalClient(og) })
  featureflag.SetGlobalClient(mymockclient)
}
```

So I'd say:
- export the `client` interface (not the global value) and the methods. 
- export the `SetGlobalClient` and `GetGlobalClient`

keiko713: 🖖

instrument/instrument.go
```diff
@@ -0,0 +1,103 @@
...
+	"gopkg.in/segmentio/analytics-go.v3"
+)
+
+type Client interface {
+	identify(userID string, traits analytics.Traits) error
+	track(userID string, event string, properties analytics.Properties) error
+	page(userID string, name string, properties analytics.Properties) error
+	group(userID string, groupID string, traits analytics.Traits) error
+	alias(previousID string, userID string) error
+}
```
rybit, line 16: if you don't export these methods then you effectively have an interface you can't implement in a different client.

keiko713: 🤦 yes

rybit (approved):

PR #229, opened 2021-04-07
feat(instrument): stop returning error with instrument functions

keiko713:
While using `instrument` in identeer, I realized that I had to do the error check every time I call the tracking.
https://github.com/netlify/identeer/pull/96/files#r602408645
This is not a user friendly shape, and also the user could only do the "logging the failure" really, nothing more. With this PR, netlify-commons will do the logging part for you, so that you can use `instrument` as a one liner.

reference: Segment would return the error with the tracking for the following case, which should be okay to "ignore":

> // The method returns an error if the message queue not be queued, which
> // happens if the client was already closed at the time the method was
> // called or if the message was malformed.

**BREAKING CHANGES:**
Functions with instrument like `instrument.Identify`, `instrument.Track` will no longer return `error`. If you were using this before with the error handling, please remove the error handling.

### DIFF
--- a/instrument/global.go
+++ b/instrument/global.go
@@ -36,26 +36,26 @@ func Init(conf Config, log logrus.FieldLogger) error {
 }
 
 // Identify sends an identify type message to a queue to be sent to Segment.
-func Identify(userID string, traits analytics.Traits) error {
-	return GetGlobalClient().Identify(userID, traits)
+func Identify(userID string, traits analytics.Traits) {
+	GetGlobalClient().Identify(userID, traits)
 }
 
 // Track sends a track type message to a queue to be sent to Segment.
-func Track(userID string, event string, properties analytics.Properties) error {
-	return GetGlobalClient().Track(userID, event, properties)
+func Track(userID string, event string, properties analytics.Properties) {
+	GetGlobalClient().Track(userID, event, properties)
 }
 
 // Page sends a page type message to a queue to be sent to Segment.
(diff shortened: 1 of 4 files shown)


rybit (commented):
Couple notes:
- should the errors be logged at Info level? I think Warning could work. 
- should we log out more information, like the userid or traits? 
- I think that if you start calling the global methods without calling `Init` you're going to panic (b/c the MockClient has a nil Logger). I think that we should not do this, it is especially annoying for tests.

keiko713:
> should the errors be logged at Info level? I think Warning could work.

yeah good point. I think the given the nature of errors, feels like the most cause is "you're doing something wrong" (client is closed early, or malformed input), so I think it can justify the warning.

> should we log out more information, like the userid or traits?

I'm having a bit mixed feeling about this; what are we trying to do with these info? Like, we want to backfill manually? I think we should be able to treat this more with "best effort reporting" than "we can't miss any message" situation.
If the error happens due to some malformed input, some info would help for debugging... are you thinking from that perspective?

> I think that if you start calling the global methods without calling Init you're going to panic (b/c the MockClient has a nil Logger). I think that we should not do this, it is especially annoying for tests.

I do have a doc explaining how you use this with test: https://github.com/netlify/netlify-commons/blob/e7030892c0d72fd6e117c4906afe1bd61e020179/instrument/doc.go#L22-L27 but yeah also nobody reads docs in reality.
I'll see what I can tweak here.

PR #236, opened 2021-05-12
feat: allow for asynchronous kafka offset commits

cbosss:
*Summary:*

Exposing `confluent-kafka-go`'s functionality for asynchronous offset commits.

This is required for https://github.com/netlify/edge-state/issues/306

### DIFF
--- a/kafka/consume.go
+++ b/kafka/consume.go
@@ -1,6 +1,5 @@
 package kafka
 
-import "C"
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


mheffner:
@pandemicsyn @kschlarman 

How are we doing this in logarchiver and logdrains? Committing every single message offset, async or not, feels like it could be expensive at scale. I wonder if there's a batching model we could share across our implementations.

pandemicsyn:
@mheffner in logarchiver we're only committing back the latest offsets in each part as we flush a large batch of messages to storage, so we end up not having to do a very high volume of commits in practice. @kschlarman also has a nice/simple Tracking package that she developed as part of other-not-public work.

PR #242, opened 2021-06-29
feat: add AllEnabledFlagsUser to featureflag

eduardoboucas:
This PR adds a `AllEnabledFlagsUser` method to the `featureflag` package, which does the same as `AllEnabledFlags` but accepts a LD user as an argument. It follows the same pattern as `Enabled`/`EnabledUser` and `Variation`/`VariationUser`.

We'd need this in buildbot as part of https://github.com/netlify/buildbot/issues/1395.


vbrown608 (approved):
:zzz:

PR #251, opened 2021-09-17
feat: added serializable URL type

cbosss:
It's common that we specify URLs in our configurations. This PR adds a serializable URL type, which can be idiomatically marshalled/unmarshalled as json or yaml. 

I also fixed a bug in our Duration type. I reworked the tests so that they now catch such regressions.

### DIFF
--- a/util/duration.go
+++ b/util/duration.go
@@ -4,8 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"time"
-
-	"gopkg.in/yaml.v3"
 )
 
 // Duration is a serializable version version of a time.Duration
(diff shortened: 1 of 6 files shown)


util/duration.go
```diff
@@ -16,8 +14,8 @@ type Duration struct {
 	time.Duration
 }
 
-func (d Duration) MarshalYAML() ([]byte, error) {
-	return yaml.Marshal(d.String())
+func (d Duration) MarshalYAML() (interface{}, error) {
```
emilyzhang, line 17: nice simplification

emilyzhang (approved):
nice!!

rybit (approved):

PR #252, opened 2021-10-08
fix: don't append to global tags from multiple goroutines

mrdg:
`permTags` is shared between (concurrent) invocations of `Distribution`, which is not safe if its `cap` is greater than `len`.


mraerino (approved):
good catch!

PR #259, opened 2021-10-31
feat: add URLParam function to router

eduardoboucas:
I wanted to define a route like `/route/{param}` and extract the value of `param` in the handler. I didn't find a built-in way of extracting the value of URL parameters with the router, so this PR adds a function that wraps the `URLParam` function from `chi`.

_Example usage:_

```go
r.Get("/jobs/{jobId}", func(w http.ResponseWriter, req *http.Request) error {
  jobId := r.URLParam(req, "jobId")
  
  doSomething(jobId)
  
  w.WriteHeader(http.StatusOK)

  return nil
})
```

Apologies if it's already possible to do this with the router and I completely missed it, in which case feel free to close the PR.

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -52,6 +52,9 @@ type Router interface {
 	Mount(pattern string, h http.Handler)
 
 	ServeHTTP(http.ResponseWriter, *http.Request)
+
+	// Returns the value of a URL parameter
+	URLParam(*http.Request, string) string
 }
 
 // New creates a router with sensible defaults (xff, request id, cors)
@@ -149,6 +152,11 @@ func (r *chiWrapper) Mount(pattern string, h http.Handler) {
 		})
 	}
 	r.chi.Mount(pattern, h)
+}
+
+// Returns the value of a URL parameter
+func (r *chiWrapper) URLParam(req *http.Request, name string) string {
+	return chi.URLParam(req, name)
 }
 
 // =======================================
--- a/router/router_test.go
+++ b/router/router_test.go
@@ -147,6 +147,21 @@ func TestVersionHeader(t *testing.T) {
 	}
 }
 
+func TestURLParam(t *testing.T) {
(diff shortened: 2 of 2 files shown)


mheffner (commented):
Cool :+1:

rybit (approved):

router/router.go
```diff
@@ -151,6 +154,11 @@ func (r *chiWrapper) Mount(pattern string, h http.Handler) {
 	r.chi.Mount(pattern, h)
 }
 
+// Returns the value of a URL parameter
+func (r *chiWrapper) URLParam(req *http.Request, name string) string {
+	return chi.URLParam(req, name)
+}
```
mraerino, line 160: i don't get why this needs to be a method on the chiWrapper?
why not make this a freestanding function?

mraerino (commented):
i've often just used `chi.URLParam` directly

router/router.go
eduardoboucas: It doesn't have to be, as it doesn't access the `chi` instance, but from a functional perspective it felt like the appropriate place. What location do you suggest? `router/helpers.go`?

eduardoboucas:
> i've often just used `chi.URLParam` directly

But that means using `chi` as a direct dependency and interacting both with the abstraction and with the underlying package, which doesn't feel ideal?

router/router.go
mraerino: i feel like it could just be some code like this to reexport it?

```go
const URLParam = chi.URLParam
```

mheffner:
Good to merge?

router/router.go
eduardoboucas: I personally prefer having a method that we explicitly implement and which we can customise the behaviour of in the future, if needed.

Is this a blocker for you?

mraerino: you can replace a reexport with a freestanding function at any point, it will not break anyone except if the signature differs which is the same if it is a function now.

PR #261, opened 2021-11-01
Add service tag to metrics and traces

mheffner:
This adds an automatic global `service` tag to metrics and traces set to the service name. This provides automatic correlation of metrics and traces emitted by a service so they can be found in DD.

We only set this if the user hasn't already provided a global `service` tag.

### DIFF
--- a/metriks/metriks_test.go
+++ b/metriks/metriks_test.go
@@ -38,11 +38,15 @@ func TestDatadogSink(t *testing.T) {
 
 	Inc("test_counter", 1)
 
-	expectedMsg := "test.test_counter:1|c|#app:edge-state,env:test"
-	buf := make([]byte, len(expectedMsg))
-	_, _, err = l.ReadFrom(buf)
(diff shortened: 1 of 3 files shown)


metriks/metriks_test.go
```diff
@@ -38,11 +38,15 @@ func TestDatadogSink(t *testing.T) {
...
+	expectedMsg := "test.test_counter:1|c|#app:edge-state,env:test,service:test"
+
+	var readBytes int
+	buf := make([]byte, 512)
+	readBytes, _, err = l.ReadFrom(buf)
 	require.NoError(t, err)
-	bytes.Contains(buf, []byte(expectedMsg))
+	require.True(t, readBytes > 0)
+
+	require.True(t, bytes.Equal(buf[0:readBytes], []byte(expectedMsg)))
```
mheffner, line 49: This test wasn't actually comparing anything as the value of `bytes.Contains` was thrown out. Also, it was truncating the msg read based on the `expectedMsg` length, so we now read up to 512 bytes.

rybit (approved):

PR #263, opened 2021-11-11
fix: panic on server shutdown from closing closed channel

kschlarman:
I've seen a few panics that seem to be caused by trying to close the `done` channel twice. This should make sure the channel only gets closed once, even if `Shutdown` is called twice.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -6,6 +6,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"os/signal"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ShaneSaww (approved):
:+1:

rybit (dismissed):

server/server.go
```diff
@@ -86,9 +88,12 @@ func New(log logrus.FieldLogger, config Config, api APIDefinition) (*Server, err
 }
 
 func (s *Server) Shutdown(to time.Duration) error {
+	s.doneOnce.Do(func() {
+		defer close(s.done)
+	})
```
mraerino, line 93: the defer is called once the function it is in ends. which in this case is immediately, so you can leave off the keywork. the concern of catching panics is not given here.

kschlarman: Ah yes, good catch. I think we still want to defer the closing of the channel though. Updated.

mraerino (approved):

Issue #255, opened 2021-10-22
`SafeRoundTripper` can leak goroutines


mrdg:
A context is created [here](https://github.com/netlify/netlify-commons/blob/master/http/http.go#L52) but the cancel function doesn't get called unless there's an error.

PR #267, opened 2021-11-17
fix: properly handle urls in saferoundtripper

### LINKED ISSUE #266: `SafeRoundtripper` assumes host string is IP
I believe this was introduced in #254. 

We assume the `hostPort` string provided to `GetConn` is an IP address, however it could be a human readable URL. 

Ex: `google.com:443`.

In this scenario, we attempt to parse the host as an IP and fail, causing the request to eventually fail. 

https://github.com/netlify/netlify-commons/blob/8cfaa4957ee236524751e92ca739c11114d2f39c/http/http.go#L63-L67

cbosss:
Closes #266. We now attempt to resolve the hostname, and check all IP addresses returned against the allow list / deny list. 

Note: `net.LookupIP` works for both URLs and IP addresses.

### DIFF
--- a/http/http.go
+++ b/http/http.go
@@ -51,9 +51,13 @@ type noLocalTransport struct {
 func (no noLocalTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	ctx, cancel := context.WithCancel(req.Context())
 
+	no.errlog.Error("CB: starting roundtrip")
+	no.errlog.Errorf("CB: allowed blocks - %v", no.allowedBlocks)
 	ctx = httptrace.WithClientTrace(ctx, &httptrace.ClientTrace{
 		GetConn: func(hostPort string) {
+			no.errlog.Errorf("CB: hostPort - %s", hostPort)
 			host, _, err := net.SplitHostPort(hostPort)
+			no.errlog.Errorf("CB: host - %s", host)
 			if err != nil {
 				cancel()
 				no.errlog.WithError(err).Error("Cancelled request due to error in address parsing")
--- a/http/http.go
+++ b/http/http.go
@@ -29,17 +29,19 @@ func init() {
 	}
 }
 
-func blocksContain(blocks []*net.IPNet, ip net.IP) bool {
+func blocksContainsAny(blocks []*net.IPNet, ips []net.IP) bool {
 	for _, block := range blocks {
-		if block.Contains(ip) {
-			return true
+		for _, ip := range ips {
+			if block.Contains(ip) {
+				return true
+			}
 		}
 	}
 	return false
 }
 
-func isPrivateIP(ip net.IP) bool {
-	return blocksContain(privateIPBlocks, ip)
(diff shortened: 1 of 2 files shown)


http/http.go
```diff
@@ -59,18 +61,19 @@ func (no noLocalTransport) RoundTrip(req *http.Request) (*http.Response, error)
 				no.errlog.WithError(err).Error("Cancelled request due to error in address parsing")
 				return
 			}
-			ip := net.ParseIP(host)
-			if ip == nil {
+
+			ips, err := net.LookupIP(host)
```
paulo, line 65: Great work! I have one question though. Does this mean will have to resolve the same host IP twice per request? (one once we get the connection and another during the normal request flow)

Been looking though how `httptrace` works and it does have a `DNSDone` hook which we could potentially use to avoid the dual lookup. The DNSDone hook will only be called when using non-idle connections (which should be fine since we're starting with a clean slate after each deploy), but it's called before actually establishing new connections, which is what we want:

- [DNSDone called here](https://cs.opensource.google/go/go/+/master:src/net/lookup.go;l=328;drc=refs%2Ftags%2Fgo1.17.3;bpv=0;bpt=0) after a DNS query
- [dns query invoked at the beginning of this block, connection created at the end](https://cs.opensource.google/go/go/+/master:src/net/dial.go;l=406-432;drc=refs%2Ftags%2Fgo1.17.2;bpv=1;bpt=1?q=dial.go&ss=go%2Fgo)

There will still be some competition on the context cancelling vs connection creation after calling `cancel()`, but I think this solution should be good enough for what we want to achieve.

I used the following playground to get the stack trace for what happens behind the scenes: https://play.golang.org/p/k_XPwRcuvMZ

mrdg: I think you're right that we do two dns lookups per request with this change, but it's likely cached most of the time? 

I think this change makes sense to fix the immediate issue, but overall it seems that `httptrace` is not really the right approach for this (there's also this issue https://github.com/netlify/netlify-commons/issues/255). I think a better approach might be to set a custom `DialContext` on the `http.Transport`. Then you can close the connection if the remote addr is not allowed, or even resolve the hostname first and return an error before you attempt a connection.

paulo: > I think you're right that we do two dns lookups per request with this change, but it's likely cached most of the time?

that depends on the OS distribution and configuration underneath, but yeah, it should be. Go itself doesn't seem to cache anything.

>  I think a better approach might be to set a custom DialContext on the http.Transport

I agree

mraerino (approved):

http/http_test.go
```diff
@@ -27,7 +27,10 @@ func TestIsPrivateIP(t *testing.T) {
 
 	for _, tt := range tests {
 		ip := net.ParseIP(tt.ip)
-		assert.Equal(t, tt.expected, isPrivateIP(ip))
+		if ip == nil {
+			require.Fail(t, "failed to parse IP")
```
rybit, line 31: I think you could use `require.NotNil`

PR #270, opened 2021-12-08
Add SafeDial to block connections

### LINKED ISSUE #255: `SafeRoundTripper` can leak goroutines

mrdg:
Closes #255

### DIFF
--- a/http/http.go
+++ b/http/http.go
@@ -88,7 +88,7 @@ func (no noLocalTransport) RoundTrip(req *http.Request) (*http.Response, error) 
 }
 
 // SafeRoundtripper blocks requests to private ip ranges
-// Deprecated: use SafeDial instead
+// Deprecated: use SafeTransport instead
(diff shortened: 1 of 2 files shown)


mraerino (commented):
could we instead alter `SafeRoundtripper` and `SafeHTTPClient` to use the SafeDial function? this would allow continued use and is probably a good abstraction right?

mrdg:
> could we instead alter `SafeRoundtripper` and `SafeHTTPClient` to use the SafeDial function? this would allow continued use and is probably a good abstraction right?

`SafeRoundtripper` takes an `http.RoundTripper` so you can't configure the dial function on that like with `http.Transport`

mraerino:
maybe offer a safe transport abstraction?

mrdg:
> maybe offer a safe transport abstraction?

Added `SafeTransport`. I kept `SafeDial` exported in case callers need to set custom dialer settings.

mraerino (approved):
nice!

PR #274, opened 2021-12-22
Report gauges once per interval

mheffner:
Gauges report a snapshot value once per interval, so we don't need to report on every change. There shouldn't be a difference between reporting N times per interval and selecting one vs. simply reporting once. This should reduce overhead of reporting for gauges with high-frequency change rates.

### DIFF
--- a/metriks/gauge.go
+++ b/metriks/gauge.go
@@ -15,9 +15,7 @@ const (
 )
 
 // PersistentGauge will report on an interval the value to the metrics collector.
-// Every call to the methods to modify the value immediately report, but if we
-// don't have a change inside the window (default 10s) after the last report
-// we will report the current value.
+//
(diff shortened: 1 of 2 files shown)


paulo (approved):
Good idea! I wonder if it would be worth it to make this behaviour configurable, so low load apps would be able to have more granular data while high load apps get sparser data without impacting perf.

mheffner:
> Good idea! I wonder if it would be worth it to make this behaviour configurable, so low load apps would be able to have more granular data while high load apps get sparser data without impacting perf.

So gauges still report once per interval, so I don't know if reporting more frequently provides any additional granularity. If an application wants to record per-operation granular metrics it is probably better to use the Sample() or Distribution() type metrics.

PR #280, opened 2022-01-03
bugsnag: Bump to latest v1

mheffner:
While there is a v2 API, at the moment logrus_bugsnag doesn't support v2.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,20 +8,20 @@ require (
 	github.com/bitly/go-simplejson v0.5.0 // indirect
 	github.com/blang/semver v3.5.1+incompatible // indirect
 	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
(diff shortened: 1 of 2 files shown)


rybit:
I was looking for any uses of that feature in our in our org, I can't find any reference to it. What do you think about deleting it? We are about to do

rybit (approved):

mheffner:
> I was looking for any uses of that feature in our in our org, I can't find any reference to it. What do you think about deleting it? We are about to do

You're referring to the `LogHook` portion?

rybit:
yeah that is what I was [searching for](https://cs.github.com/?scopeName=netlify&scope=org%3Anetlify&q=log_hook%20org%3Anetlify) it would be specified in the config files.

mheffner:
yeah, looks like @mraerino has a parallel PR in #281 so happy to close this one

mraerino:
sorry didn't see this was open

PR #304, opened 2022-01-11
chore: Aquire token for release please on the fly

mraerino:
Release-please on main is broken right now, since the Github API token i tried to use in https://github.com/netlify/netlify-commons/pull/301 should not be used anymore.

This new method acquires a token from a github app instead which is much nicer.


mheffner (approved):
cool

PR #313, opened 2022-01-21
chore: update how we get request IDs

rybit:
I was debugging some other service I'm starting and noticed that I wanted to validate that (1) the tracer is present and (2) that we're doing the start/finished lines. That made me realize that we're not copying the hooks over. So I fixed that. 

Then I wanted to improve a few things:
- make the tracer package always return a request ID so we can use that to generate missing one
- have a reliable way to get a per-request logger. So export that function.

### DIFF
--- a/router/router_test.go
+++ b/router/router_test.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


mraerino (approved):
cool

PR #321, opened 2022-01-27
feat: added serializable http.Header type

cbosss:
It's common that we specify Headers in our yaml/json configurations. This PR adds a serializable http.Header type, which can be idiomatically marshalled/unmarshalled.

### DIFF
--- a/util/headers.go
+++ b/util/headers.go
@@ -1,0 +1,48 @@
+package util
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// Headers is a serializable version of http.Header it supports both yaml & json formats.
+// Headers expects the yaml/json representation to be a map[string]string.
+type Headers struct {
+	http.Header
+}
+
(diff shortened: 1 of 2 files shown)


mraerino (approved):
i guess this could be optimized for perf (to allocate less), but looks good to me for now.

mraerino:
> Just curious... how would you optimize for less allocs here? Is it possible to incrementally read a json map in?

yeah you can use `json.Decoder` to read tokens and call the `Add` directly without the intermediate map

cbosss:
Cool! TIL about the `Token()` call on `json.Decoder`. 

I played around with it a little bit, and I think it will add more complexity than it's worth. I envision this code mainly being used for parsing config files, so it's unlikely there will be large maps in there, nor will we be repeatedly parsing the file. I think i'm just going to ship this simpler / less performant version. 

Thanks for the info though!

PR #325, opened 2022-02-15
DNMY - figure out how to get reporting for endpoints with the path

rybit:
ok this is not going to be merged as is, more as a discussion. 

I was working on some other service and realized I wasn't logging 401s and 404s. So I traced that back to the router. We never get a call for a 404, duuuuh I went. So I found that we have handlers for not found and method not allowed. All is good in the world.

BUT

No. So I was looking and if your middleware stack rejects the call _before_ we get to the actual handler, so think CORS and auth, we won't log/count either. The problem is that we only wrap when we get to the very end. Then naturally we need to tweak the middleware stack to be _first_ and then let the other handlers reject as makes sense. Sweet. Easy. 

BUT 

We try to encode the 'pattern' in the handler's APM for better visibility. This way we can get top level paths in datadog for `GET.my.special.path` and not for every unique path at the top. This is where we are now. The problem is that (1) chi doesn't know the path until _after_ it has invoked the handler. That is _after_ we could block the call. (2) we can't really 'register' any object I can think of or put it in the context to let us know that until after the route is defined. This means I can't easily see how often `GET.my.authenticated.endpoint` is getting hit with a 401. Yes we can log that now (yay) but spans won't show that. 

What I'm thinking of doing is actually maybe splitting the problem. Leave the code ~as is (just the err handlers). But also start a parent span we report if we don't report another? IDK I'm pretty stumped and don't think that will actually work. Or maybe just start a new child span and put the pattern there?

### DIFF
--- a/router/middleware.go
+++ b/router/middleware.go
@@ -147,6 +147,12 @@ func BugSnag() Middleware {
 	})
 }
 
+type tracingMiddleware struct {
+	patterns map[http.Handler]string
+	log      logrus.FieldLogger
+	svcName  string
+}
+
 func TrackAllRequests(log logrus.FieldLogger, service string) Middleware {
 	return MiddlewareFunc(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
 		// This is to maintain some legacy span work. It will cause the APM requests
--- a/router/options.go
+++ b/router/options.go
@@ -1,37 +1,81 @@
 package router
 
-type Option func(r *chiWrapper)
+import (
+	"net/http"
 
-func OptEnableCORS(r *chiWrapper) {
-	r.enableCORS = true
+	"github.com/netlify/netlify-commons/tracing"
+)
+
+type Option struct {
+	f func(r *chiWrapper)
+}
+
+func OptEnableCORS() Option {
+	return Option{
+		f: func(r *chiWrapper) {
+			r.enableCORS = true
+		},
+	}
 }
 
 func OptHealthCheck(path string, checker APIHandler) Option {
-	return func(r *chiWrapper) {
-		r.healthEndpoint = path
-		r.healthHandler = checker
+	return Option{
+		f: func(r *chiWrapper) {
+			r.healthEndpoint = path
+			r.healthHandler = checker
+		},
 	}
 }
 
 func OptVersionHeader(svcName, version string) Option {
-	return func(r *chiWrapper) {
-		if version == "" {
-			version = "unknown"
-		}
-		r.version = version
-		r.svcName = svcName
+	return Option{
+		f: func(r *chiWrapper) {
+			if version == "" {
+				version = "unknown"
+			}
+			r.version = version
+			r.svcName = svcName
+		},
 	}
 }
 
 func OptEnableTracing(svcName string) Option {
-	return func(r *chiWrapper) {
-		r.svcName = svcName
-		r.enableTracing = true
+	return Option{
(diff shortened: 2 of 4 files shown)


router/middleware.go
```diff
@@ -147,6 +147,12 @@ func BugSnag() Middleware {
 	})
 }
 
+type tracingMiddleware struct {
+	patterns map[http.Handler]string
```
rybit, line 151: this won't work (I was trying to bend go here). Can't index on a function

router/options.go
```diff
@@ -1,37 +1,81 @@
...
+	"net/http"
 
-func OptEnableCORS(r *chiWrapper) {
-	r.enableCORS = true
+	"github.com/netlify/netlify-commons/tracing"
+)
+
+type Option struct {
+	f func(r *chiWrapper)
+}
```
rybit, line 11: This is a change I'd actually keep, exposing the `Option` but not the `chiWrapper` is kinda..odd. And as it didn't work in the past b/c `chiWrapper` wasn't exposed, this makes it a safe breaking change (and will be doc'd that way).

router/router.go
```diff
@@ -107,7 +115,12 @@ func (r *chiWrapper) Method(method, pattern string, h APIHandler) {
 
 // Get adds a GET route
 func (r *chiWrapper) Get(pattern string, fn APIHandler) {
-	r.chi.Get(pattern, r.traceRequest(http.MethodGet, pattern, fn))
+
+	wrapper := new(chiWrapper)
+	*wrapper = *r
+	wrapper.tracingPrefix = sanitizePattern(pattern)
+
+	wrapper.chi.Get(pattern, wrapper.traceRequest(http.MethodGet, pattern, fn))
```
rybit, line 123: this won't work because the `tracingPrefix` is associated with the root `chi` router. Though I'm not totally convinced that I can't get this to work...just couldn't at the time.

router/middleware.go
mraerino: i think if you make it an `unsafe.Pointer` it would work, but maybe there is a better way

router/options.go
rybit: Mostly it is about the exposed signature. It is an exported type `Option` which has a signature using a non exported type `chiWrapper` which I think is messy. Functionally, not much changes as you can see

mraerino:
could you use reflection to get the name of the function being registered? 😈

rybit:
The problem is that the calls look like this:
```
  router.Route
    healthMiddleware
    .... whatever ...
    I reject things middleware
    your route handler
```

So if you get to the reject Middleware, you never get to the route handler. That handler is the one that currently triggers tracing and has access to the pattern (or any other marker we could). What we'd need to do to capture middleware level rejects is change it to:
```
router.Route
  healthMiddleware
  tracingMiddleware <-- note at the top of the chain
  ... the rest of the stuff ...
```

When we're at the tracing middleware, we don't know what the eventual handler is going to be. So we can't use the information that is down that low. `chi` has a `RoutePattern` field, but it is populated _after_ the request is complete. So we can't use it on the starter span. 

I'm trying to think if this is something we can tweak to more of a builder pattern (it's not far from it) so that we could record the path, but honestly it isn't solving the problem of resolving at runtime.

mraerino:
maybe this helps? https://github.com/go-chi/chi/issues/270#issuecomment-479184559

rybit:
related: https://github.com/netlify/identeer/issues/53

PR #326, opened 2022-02-24
feat: HTTPError.WithFields

Skn0tt:
This PR adds `.WithFields` and `.WithField` methods to `HTTPError`, which act as a facade to Logrus' methods.
They allow to easily add custom log fields to error messages.

# how this helps

At the moment, when you want to add additional information to an HTTP error, you need to print two log lines:

```go
logger.WithError(err).WithField("body", string(body)).Error("error unmarshaling events")
return router.BadRequestError("error parsing body: %s", err.Error())
```

With this PR, we can trim it down to one:

```go
return router.BadRequestError("error parsing body: %s", err.Error()).WithInternalError(err).WithField("body", string(body))
```

(taken from https://github.com/netlify/ingesteer/pull/18#discussion_r813652764)

### DIFF
--- a/router/errors.go
+++ b/router/errors.go
@@ -8,6 +8,7 @@ import (
 	"github.com/bugsnag/bugsnag-go/v2"
 	"github.com/netlify/netlify-commons/metriks"
 	"github.com/netlify/netlify-commons/tracing"
+	"github.com/sirupsen/logrus"
 )
 
 // HTTPError is an error with a message and an HTTP status code.
@@ -18,6 +19,7 @@ type HTTPError struct {
 	InternalError   error       `json:"-"`
 	InternalMessage string      `json:"-"`
 	ErrorID         string      `json:"error_id,omitempty"`
+	Fields          []logrus.Fields
 }
 
 // BadRequestError creates a 400 HTTP error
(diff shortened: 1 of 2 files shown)


router/errors.go
```diff
@@ -8,16 +8,18 @@ import (
...
-	InternalError   error       `json:"-"`
-	InternalMessage string      `json:"-"`
-	ErrorID         string      `json:"error_id,omitempty"`
+	Code            int             `json:"code"`
+	Message         string          `json:"msg"`
+	JSON            interface{}     `json:"json,omitempty"`
+	InternalError   error           `json:"-"`
+	InternalMessage string          `json:"-"`
+	ErrorID         string          `json:"error_id,omitempty"`
+	Fields          []logrus.Fields `json:"-"`
```
mraerino, line 22: `logrus.Fields` is just an alias for a hashmap. multiple entries should be merged, which is the same logrus does for multiple calls to `WithFields`
you could also use a `*logrus.Entry` which you could call `WithFields` on to use their merging logic

Skn0tt: I added some naive merging logic, and a test: https://github.com/netlify/netlify-commons/pull/326/commits/b692cb1e06019e100b57897652c251e3c4627ff3

mraerino (approved):

PR #327, opened 2022-03-08
chore: Add Hostname to Bugsnag Configuration

aarthykc:
## Context
This PR adds the hostname to the Bugsnag Configuration for better debugging. 

## Next Steps
- [x] Receive at least 1 recent approving review 
- [ ] Release a new version

### DIFF
--- a/nconf/bugsnag.go
+++ b/nconf/bugsnag.go
@@ -8,7 +8,7 @@ type BugSnagConfig struct {
 	Environment    string
 	APIKey         string `envconfig:"api_key" json:"api_key" yaml:"api_key"`
 	ProjectPackage string `envconfig:"project_package" json:"project_package" yaml:"project_package"`
-	CDNHostName    string `envconfig:"node_name" json:"node_name" yaml:"node_name"`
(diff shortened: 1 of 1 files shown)


nconf/bugsnag.go
```diff
@@ -8,6 +8,7 @@ type BugSnagConfig struct {
 	Environment    string
 	APIKey         string `envconfig:"api_key" json:"api_key" yaml:"api_key"`
 	ProjectPackage string `envconfig:"project_package" json:"project_package" yaml:"project_package"`
+	HostNodeName   string `envconfig:"node_name" json:"node_name" yaml:"node_name"`
```
cbosss, line 11: Can you add a comment here... something like: 

`If left unset, bugsnag will default to the value returned by os.Hostname`.

mraerino, line 11: why not call the field just `NodeName`?

mraerino (commented):
very nitty comment. code looks good.

nconf/bugsnag.go
aarthykc: Donezo!

aarthykc: Donezo!

cbosss (approved):

PR #336, opened 2022-05-19
chore: remove the parser package to its own repo

rybit:
In an effort to reduce the blast radius of this repo, I've broken out the parser to their own repository: https://github.com/netlify/go-parsers

cc @kschlarman @ShaneSaww 

BREAKING CHANGES:
This removes the entire parser package


ShaneSaww (approved):
The first commit of https://github.com/netlify/go-parsers looks like it matches 
so
lgtm

PR #338, opened 2022-06-21
feat: make Router.enableTracing configurable

Skn0tt:
This PR adds a `config.Router.EnableTracing` field that makes `Router.enableTracing` configurable.

Works towards https://github.com/netlify/ingesteer/issues/152.

There are no breaking changes introduced in this.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -32,7 +32,7 @@ type Server struct {
 }
 
 type RouterConfig struct {
-	EnableTracing bool
+	DisableTracing bool
 }
 
 type Config struct {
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -31,11 +31,16 @@ type Server struct {
 	doneOnce sync.Once
 }
 
+type RouterConfig struct {
+	EnableTracing bool
```
rybit, line 35: So I think that this would be oddly breaking in a functional way (the code would work). Updating here would _disable_ this by default. I think that we should consider flipping the boolean here (e.g. `DisableTracing`).

Skn0tt: ooooh good point

Skn0tt: done in https://github.com/netlify/netlify-commons/pull/338/commits/2351a826b214c4fe13f74106833796cf81d450c6

rybit (approved):

PR #340, opened 2022-06-24
chore: in release-please PRs, at-mention participating authors

Skn0tt:
This PR adds a Github Action that scrapes `CHANGELOG.md` for commits part of the release, finds the respective authors, and tags them on the release. I'm hoping this helps against release-please PRs idling around, as authors will now receive notifications about them.

If this proves useful, we can abstract it and add it to other repositories as well.


.github/workflows/release-please-mention-contributors.yml
```diff
@@ -0,0 +1,54 @@
...
+            const authors = []
+            for (const match of diff.matchAll(COMMIT_URL_REGEX)) {
+              const ref = match[1]
+              const [owner, repo] = "${{ github.repository }}".split("/")
+              const commit = await github.rest.repos.getCommit({ owner, repo, ref })
+              const { author } = commit.data
+              if (author) {
+                authors.push(author.login)
+              }
+            }
```
mraerino, line 33: maybe rewrite this `for` into a `Promise.all` with `map`?

Skn0tt: not all commits have an author (bots, for example, don't), so this would require `flatMap`, which would make things harder to understand. I think it's fine to use a `for` here.

mraerino: you'd just need `map(...).filter()` - not really hard to understand

Skn0tt: implemented in f22d5ceeea838a02ce7b44cce6a785eb4fbbc106

PR #345, opened 2022-07-11
chore: add removal notice

rybit:
This is to provide notice that these different packages are actively not being maintained and migrated away from. I'd like to remove them but there is a bit of a rats nest of dependencies and until we can move the server/http/router packages (and make them depend on go-*) then it is more work to strip them out. 

So as a stop gap until I get more time to work on this (next first friday?), I wanted to put a notice in the README.


emilyzhang (approved):
love the useful documentation!

PR #346, opened 2022-08-10
feat: allow disabling trace logging, while keeping Datadog tracing enabled

Skn0tt:
This PR extends the `DisableTracing` config property added in https://github.com/netlify/netlify-commons/pull/338 to add a more fine-granular `DisableTraceLogging` option.

This allows services like Ingesteer, which have lots of requests but don't really need all of them logged, to only enable Datadog tracing, but not log information about every single request to Humio.

Works towards https://github.com/netlify/ingesteer/issues/169.

### DIFF
--- a/router/errors_test.go
+++ b/router/errors_test.go
@@ -26,6 +26,7 @@ func TestHandleError_ErrorIsNil(t *testing.T) {
 		logger,
 		"test",
 		"test",
+		true,
 	)
 
 	HandleError(nil, w, r)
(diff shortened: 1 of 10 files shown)


mraerino (commented):
`tracing.TrackRequest` is a public api so this is a breaking change?

also: we're moving to https://github.com/netlify/go-observability and other libs for our services to break up the commons lib. maybe you want to take this opportunity to move ingesteer over and make the change there instead?

Skn0tt:
technically, it's public, in reality, it's used by nobody: https://cs.github.com/?q=owner%3Anetlify+TrackRequest&p=1&pt=3ef9cebfa2b06556f3d994c753e4f3781cc51ab41b7f91c3ee5be7937f355c30f345ec0f2cd12b90e3716c62eaca4e8341f081aecdd4f5b12163a2b19960f7ef20c448ebad0656537f074aca7c3fc2969f0f82ccc5eba852169382d7&scope=&scopeName=All+repos

happy to move to go-observability though, sounds good!

Issue #347, opened 2022-10-06
chore: support netlify.toml and deprecate netlify.json and netlify.yaml files


4xposed:
When looking into https://github.com/netlify/buildbot/issues/2141 we saw that files in `yaml`, `yml` and `json` are still supported in `netlify-commons/ntoml` package.

But in `build` we only make use of `netlify.toml` https://github.com/netlify/build/blob/60523200de880d86716acb6079e2699e32cd90f9/packages/config/src/path.js#L52

This ticket is to remove all logic related to support the different .yaml/.yml/.json files

PR #348, opened 2022-10-06
chore: remove support for netlify.json/.yaml/.yml

4xposed:
Addresses: https://github.com/netlify/netlify-commons/issues/347

When looking into https://github.com/netlify/buildbot/issues/2141 we saw that files in `yaml`, `yml` and `json` are still supported in `netlify-commons/ntoml` package.

But in `build` we only make use of `netlify.toml` https://github.com/netlify/build/blob/60523200de880d86716acb6079e2699e32cd90f9/packages/config/src/path.js#L52

This PR removes all logic related to support the different .yaml/.yml/.json files

### DIFF
--- a/ntoml/netlify_toml.go
+++ b/ntoml/netlify_toml.go
@@ -1,55 +1,47 @@
 package ntoml
 
 import (
-	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
-	"strings"
 
 	"github.com/BurntSushi/toml"
 	"github.com/pkg/errors"
-	yaml "gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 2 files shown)


ntoml/netlify_toml.go
```diff
@@ -148,12 +115,6 @@ func LoadFrom(paths ...string) (*NetlifyToml, error) {
 			switch extension {
 			case ".toml":
```
mraerino, line 116: could you maybe simplify this code since we don't really care about different extensions anymore?

mraerino (commented):
mostly good, one thing

ntoml/netlify_toml.go
4xposed: done 👍

mraerino (approved):
LGTM

PR #354, opened 2022-10-07
remove packages that have been broken out

rybit:
BREAKING CHANGES:
- removes the packages:
  - `metriks` -> github.com/netlify/go-observability
  - `tracing` -> github.com/netlify/go-observability
  - `utils` -> github.com/netlify/go-utils
  - `testutils` -> github.com/netlify/go-test-utils
  - `server` -> github.com/netlify/go-http
  - `router` -> github.com/netlify/go-http
  - `featureflag` -> github.com/netlify/go-flags
  - `nconf` -> github.com/netlify/go-config

Then in each service we'd be able to configure those utilities in the service to allow for independent versioning per service. No more huge upgrades! 

Note: these are all private packages, we need to make sure that you've got CI setup to access them properly

Examples:
- https://github.com/netlify/go-service-boilerplate/pull/74
- https://github.com/netlify/stargate/pull/685
- https://github.com/netlify/dos-center/pull/234


mongoclient/tls.go
```diff
@@ -0,0 +1,100 @@
+package mongoclient
```
rybit, line 1: this is stolen from go-http/ntls

paulo (approved):

